PR #24, opened 2015-11-04
Update MovieSelection.py

digicasi:
don't stop live preview when exiting movie list.

### DIFF
--- a/lib/python/Screens/MovieSelection.py
+++ b/lib/python/Screens/MovieSelection.py
@@ -1394,11 +1394,13 @@ class MovieSelection(Screen, HelpableScreen, SelectionEventInfo, InfoBarBase, Pr
 			return
 
 		self.saveconfig()
-		from Screens.InfoBar import InfoBar
-		infobar = InfoBar.instance
-		if self.session.nav.getCurrentlyPlayingServiceReference():
-			if not infobar.timeshiftEnabled() and ':0:/' not in self.session.nav.getCurrentlyPlayingServiceReference().toString():
-				self.session.nav.stopService()
+		
+		if not config.movielist.show_live_tv_in_movielist.value:
+			from Screens.InfoBar import InfoBar
+			infobar = InfoBar.instance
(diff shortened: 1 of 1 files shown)


RobvanderDoes:
I tested this, but I can't see any difference in behaviour. Please explain what you mean exactly.

digicasi:
hi rob, 
here's the procedure to see the difference:
- enter the movie/recordings list; life tv channels continues to play in the pig frame (if enabled in settings)
- when you exit the movie list life tv gets disrupted (black frame)
- with the fix on exit live tv channel continues without disruption in full size.
this is similar to the epg behavior.
thanks for checking
digicasi aka 2stein

digicasi:
sorry, erroneously close the merge request

RobvanderDoes:
That's what I thought it would do, but it doesn't. Behaviour is identical.
But this may be dependant on hardware/drivers: I tested on GB Quad +.

digicasi:
I'm on VU+ DUO2.

digicasi:
could anyone with a duo2 please test this? works here nicely. :-) thanks.

RobvanderDoes:
As it doesn't work for the Quad as I wrote above, there's not much point in testing on another box. We can't accept changes for a specific box.

andyblac:
the black frame as you call it, is needed so the channel can restart PTS.

digicasi:
ok... but if i just enter the movie list (without starting a movie playback) and then exit again... there is no need to restart pts, right? in this case my if statement may need to be a bit more specific (might need your help here) thanks.

andyblac:
you could something to check for a preview started or a movie played, and if TRUE then only do the service stop.

digicasi:
yes, that's an option. however, i wonder whether the stopService() is really required in abort() because it is already done in playbackStop().

digicasi:
don't think i can do better than that.
works perfectly on vu+ and shouldn't hurt on other boxes.

digicasi:
forget it.

PR #35, opened 2016-05-02
Add option of customising the coloured key options within GEPG

bbbuk:
Thanks to ketmp/darkdreamdan/matt007/bbbuk


AbuBaniaz:
@bbbuk 
Might be an idea to close this request. Simple reason is that it is not wanted by some members of the team. This is despite that defaults are existing.

I was asked to test it after it was suspected that it would slow down the EPG. I have tested the modifications and not found an issue with it. The changes as part of trashcan sort are easy to work around.

rossi2000:
wait for new mod.

PR #43, opened 2016-10-09
add guidebouquetlist for supported skins

davesayers2014:
guidebouquetlist should only display if supported by the skin and shouldn't effect skins that dont support them

### DIFF
--- a/lib/python/Screens/EpgSelection.py
+++ b/lib/python/Screens/EpgSelection.py
@@ -87,6 +87,7 @@ class EPGSelection(Screen, HelpableScreen):
 		self.NumberZapField = None
 		self.CurrBouquet = None
 		self.CurrService = None
+		self["guidebouquetlist"] = Label()
 		self["number"] = Label()
 		self["number"].hide()
 		self['Service'] = ServiceEvent()
@@ -111,7 +112,7 @@ class EPGSelection(Screen, HelpableScreen):
 				'OKLong': (self.OKLong, _('Zap to channel and close (setup in menu)'))
 			}, -1)
 		self['okactions'].csel = self
-		self['colouractions'] = HelpableActionMap(self, 'ColorActions', 
+		self['colouractions'] = HelpableActionMap(self, 'ColorActions',
 			{
 				'red': (self.redButtonPressed, _('IMDB search for current event')),
 				'redlong': (self.redButtonPressedLong, _('Sort EPG List')),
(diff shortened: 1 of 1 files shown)


davesayers2014:
You can scroll through them by pressing previous and next to scroll between bouquets this part of the code was used in the OpenVix Custom EPG plugin this needs supporting by the skin designer http://www.world-of-satellite.com/attachment.php?attachmentid=49215&d=1470035421

RobvanderDoes:
ad 1): You didn't understand my question. Changing bouquets via Next/Previous was already available in all EPG's. My question was about the screen widget 'guidebouquetlist': it shows a list of all the bouquets, the actual one being the first to be displayed. What's the purpose of this, when you can't scroll through that list (the widget)? I could understand the (help) functionality if it would only display the name of the Next/Previous bouquet, but that's not the case in your code.

ad 2): You didn't answer that question.  Your code adds the widget 'guidebouquetlist' to _ALL_ EPG's, but it's empty (apart from the one in the graphical EPG).

davesayers2014:
1) you could see what bouquets are further along the list so you know how many times to keep pressing next to get to the bouquet you want at the moment you can't evan see whats next.

2) I'm not sure I just used winmerge to add the code from an alternative epgselection.py maybe someone else can add support for the rest of the screens

RobvanderDoes:
ad 1): But as I tried to explain: the list only shows the actual bouquet and the next bouquet(s), never the previous one(s). So that makes it rather useless IMHO. If it would show the previous & the next bouquet(s) I can understand the usefulness of this widget.

ad 2): As it is now we can't accept this code. We always did and do our utmost to make the UI of all EPG's the same. This codes breaks that, and adds unused code to all other EPG's.

davesayers2014:
Ok thank you for looking into it I'll close the pull request and look into your suggestions for having previous and next bouquets and adding it to other epg screens

RobvanderDoes:
Good, and thanks for that. If those two issues would be fixed, this seems to be a useful enhancement.

RobvanderDoes:
Just to be sure I made myself clear, and to prevent unneeded work for you: in your code the 'guidebouquetlist' is already present in all EPG's. But only in the graphical EPG it's actually filled with text (the bouquet names), in the other EPG's it's empty.

AbuBaniaz:
If you do submit another pull request, it should be against the Dev branch. Would help testing if you can say which skin use the feature

RobvanderDoes:
For testing purposes you can simply add the widget to any skin. No need to adjust lay-out, just see if/how it works.
Use something like:
`<widget name="guidebouquetlist" position="xxx,yyy" size="aaa,bbb" font="Regular;33"/>`

PR #48, opened 2016-11-16
Fix ConfigList.py to run onDeselect/onSelect only if

original-birdman:
only if self.current != self.getCurrent()
i.e. if the selection has *actually* changed...
This means that Notifiers with immediate_feedback = False actually
do only get called once at the end of an item change, not for every
step along the way.


original-birdman:
I've added the "Are you sure?" changes too (I thought they would be kept separate, but apparently not).

PR #76, opened 2017-01-02
Allow use of font aliases in TemplatedMultiContent

IanSav:
This change allows skin developers to use font aliases as well as normal fonts in skin based TemplatedMultiContent converter structures.

In the past users could only specify fonts in a TemplatedMultiContent converter structure by using the " "fonts": [gFont("Regular",28)] " syntax.  With this change users can now also specify font aliases defined in the < fonts > section of the skin as well as the aliases hard coded in skin.py.  For example, " "fonts": [parseFont("Body;25")] ".  This change should make maintaining skin font management a much easier process.


IanSav:
Is there a problem with this pull request?

It has been closed (rejected) rather than merged and there is no reason for the closure.

RobvanderDoes:
It has been accepted, see https://github.com/OpenViX/enigma2/commit/acfe6936e75503e3ae11b4c9397149cd7e9fc555

IanSav:
Github sent me a "Closed" notification and not the usual "Merged" notification.

PR #82, opened 2017-01-05
Use new date and time formatting options

IanSav:
The changes to the various modules in this pull request are to remove the hard coded date and time formats and use the new user selectable date / time formats.

Please note that as a result of the new global configuration options for date and time the specific time settings for the EPG timeline are no longer required.  The timeline format is now synchronised to global time format. This change eliminates the need for the config.epgselection.graph_timeline24h and config.epgselection.infobar_timeline24h configuration settings.

[setup.xml] Remove deprecated EPG timeline 24hour configurations

[UsageConfig.py] Remove deprecated EPG timeline 24hour configurations


IanSav:
This request is being closed to clean up and create a new branch for the change.

PR #90, opened 2017-01-07
Enhance EventName and genre

IanSav:
These changes are to implement the following:

- Add the infrastructure to localise the ratings and genres.  The current "ETSI" standard in maintained as the default but developers can add other standards as required.  To start things off I have added the structures for Australia ("AUS").  Other countries with different rules can be similarly added.

    The switch between versions of rating and genre are automatically triggered by the country coded in the transmission.  If required a setting option could be added to override this automatic selection and force a particular standard.

- Add options to the EventName converter to allow skin writers finer control on how the items offered by this converter are displayed.  The arguments and options are provided in a comma separated list.  (Given that some builds use commas and others semicolons to separate converter parameters this code will accept both!)

    - Option "Separated" places a blank line between the description and extended description in "FullDescription" displays.
    - Option "NotSeparated" places the extended description on the following line after the description in "FullDescription" displays.  This is the default.
    - Option "Trimmed" strips leading and trailing spaces from the item being displayed.
    - Option "NotTrimmed" shows the item as defined.  This is the default.

    The options have enable and disable values to allow the skin developer to specify the display they require that do not depend on the default defined in the code.

- Refactor the code to simplify the structure of the code while allowing the additional functionality.

Some of these changes are based on the efforts of Prl in the Beyonwiz repository.


RobvanderDoes:
Accepted to make it easier for all to test.

PR #93, opened 2017-01-08
[EventName] Correct calculation of PDC time

IanSav:


IanSav:
Too many mistakes, I will resubmit with a clean history.

PR #104, opened 2017-01-14
Remove positioner test pattern on exit or cancel

IanSav:
This change stops the current service when loading the screen position test pattern and restarts the service when the settings are saved or the operation is cancelled.

Also perform a PEP8 cleanup.


Huevos:
Please do not merge this commit. Contains unnecessary formatting changes that will cause merge conflicts in the future.

PR #109, opened 2017-01-15
EPG Enhancements

IanSav:
- Restore manual EPG timeline configuration.  This change restores the ability to separately configure the timeline clock display (between 12 and 24 hour times). The settings defined are remembered but ignored if the system wide time settings are available.  The settings selected will be used if the current skin does not allow for the system wide time settings.

- Add an advisory note to the two EPG timeline setting options to warn users that these settings will be ignored if the current skin uses the system wide time formats.

- The MultiEPG column width code has been simplified.  There is no functional change.

- The SingleEPG column width code has been refined to improved the skin based parameter overrides.  The date/time column is now specified as the desired width assuming a 24 hour clock and will be automatically expanded for 12 hour clocks.  The description column is now specified as an offset from the end of dynamic date/time column rather than a hard starting point.  This change reduces wasted space between the date/time and description columns.

- Add an advisory note to the two EPG timeline setting options to warn users that these settings will be ignored if the current skin uses the system wide time formats.

### DIFF
--- a/data/setup.xml
+++ b/data/setup.xml
@@ -328,7 +328,7 @@
 		<item level="2" text="Picon width" description="Configure the width allocated to the picon.">config.epgselection.infobar_piconwidth</item>
 		<item level="2" text="Info icon width" description="Configure the minimum width before showing the 'i' icon, '0' is equal to not showing the icon at all.">config.epgselection.infobar_infowidth</item>
 		<item level="2" text="Timeline font size" description="Configure the font size relative to skin size, so 1 increases by 1 point size, and -1 decreases by 1 point size.">config.epgselection.infobar_timelinefs</item>
-		<item level="2" text="Timeline 24 Hour" description="Show time in 24 hour clock format.">config.epgselection.infobar_timeline24h</item>
+		<item level="2" text="Timeline 24 Hour" description="Show time in 24 hour clock format.\nNOTE: This setting is not used if the global time format settings are available.">config.epgselection.infobar_timeline24h</item>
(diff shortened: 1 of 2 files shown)


IanSav:
In the latest version of this code the four EPG parameter lists are:

`<parameter name="EPGMultiColumnScales" value="6.5,6.0,4.5,1.5" />`

The values are:
- the scale width of the service column (6.5)
- the scale width of the progress/start/end column (6.0)
- the scale width of the duration column (4.5)
- the scale width modifier of the progress/start/end column for 12 hour clock times (1.5)
 
`<parameter name="EPGMultiColumnSpecs" value="0,153,10,149,18,10,120,10" />`

The values are:
- the pixel offset of the start of the service column (0)
- the pixel width of the service column (153)
- the pixel width of the separation between the service and progress/start/end columns (10)
- the pixel width of the progress/start/end column (149)
- the pixel height of the progress bar (18)
- the pixel width of the separation between the progress/start/end and duration columns (10)
- the pixel width of the duration column (120)
- the pixel width of the separation between the duration and  event name columns (10)
 
`<parameter name="EPGSingleColumnScales" value="5.7,6.0,1.5" />`

The values are:
- the scale width of the date column (5.7)
- the scale width of the start/end column (6.0)
- the scale width modifier of the start/end column for 12 hour clock times (1.5)
 
`<parameter name="EPGSingleColumnSpecs" value="0,125,5,135,20" />`

The values are:
- the pixel offset of the start of the date column (0)
- the pixel width of the date column (125)
- the pixel width of the separation between the date and start/end columns (5)
- the pixel width of the start/end column (135)
- the pixel width of the separation between the start/end and event name columns (20)

PR #115, opened 2017-01-17
[UserInterfacePositioner] Improve UI/functonality

IanSav:
- Stop the running service just before the test pattern is loaded and restore it service when the user exits
- Add an usage / information pop up option to the BLUE button
- Change the order of the settings to be more logical
- Promptly update the screen when YELLOW (Default) is pressed
- Update all the on-screen values when YELLOW (Default) is pressed
- Delete the unused "run" function


IanSav:
The changes are to be reworked.

PR #141, opened 2017-03-05
Set the default debugTime to true, as logs without this lose a lot of…

original-birdman:
… relevance.

My debug logs on OpenViX 5.0 are now just a bunch of text lines with no way to relate consecutive lines at all. They may be related, or they may be many minutes apart. Useless.


original-birdman:
Why has this been closed?
Why have the debug logs been left in a state such that they are useless?

PR #146, opened 2017-03-19
Do not use Europe/Istanbul for non-Istanbul entires

original-birdman:
Places such as Helsinki have had to set Asia/Jerusalem to get the correct time, but that goes on to DLST on Friday Mar 24th, so it will be wrong then to.
This changes "Helsinki, Kyiv, Sofia" to use Europe/Helsinki, "Riga, Tallinn, Vilnius" to use Europe/Riga and "Bucharest" to use Europe/Bucharest.  "Athens, Istanbul, Minsk" is left as Europe/Istanbul, so Athens and Minsk users will have to use one of the others.

NOTE: that this really needs to get into production by the end of this week!!

### DIFF
--- a/data/timezone.xml
+++ b/data/timezone.xml
@@ -37,10 +37,10 @@
 	<zone name="(GMT+01:00) West Central Africa" zone="Africa/Kinshasa" />
 	<zone name="(GMT+01:00) Algiers, Tunis" zone="Africa/Kinshasa" />
 	<zone name="(GMT+02:00) Athens, Istanbul, Minsk" zone="Europe/Istanbul" />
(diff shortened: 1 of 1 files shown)


sunder67:
NOTE: Turkey time zone fixed at UTC +3
https://www.timeanddate.com/news/time/turkey-scraps-dst-2016.html

original-birdman:
Yes, I know.
The problem with Vix (and OE in general) is that it's based on an MS Windows view for setting timezones, which is wrong.

I'm working on code to let you set the real timezone from the zoneinfo files on the system. This works OK, unless you cancel the setting, in which case the menu code never sends a callback about the cancel.

This timezone file is juts a workaround until I can get the menu callback to work sensibly.

PR #149, opened 2017-03-22
Extend the functionality of Notifiers

original-birdman:
This extends the functionality of ConfigElement notifiers by:

a) adding the ability to have callbacks when menus are Saved and Cancelled to FinalNotifiers (it already exists for non-Final ones).

b) Adding a _callback_reason to the callback ConfigElement during any initial call ('initial'), Save ('save') or Cancel ('cancel') action. This attribute is only set for the duration of the call and is not set for any  normal or user-activated call.

This enables callback function to initialize items and also to only perform configuration changes when the final value has been chosen (or do nothing if the actin is discarded).

This is required for the correct functioning of a two-item timezones menu (area then timezone in area) that will follow.

### DIFF
--- a/lib/python/Components/config.py
+++ b/lib/python/Components/config.py
@@ -38,6 +38,7 @@ class ConfigElement(object):
 		self.__notifiers_final = None
 		self.enabled = True
 		self.callNotifiersOnSaveAndCancel = False
+		self.callFinalNotifiersOnSaveAndCancel = False
 
 	def getNotifiers(self):
 		if self.__notifiers is None:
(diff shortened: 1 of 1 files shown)


AbuBaniaz:
Reason for closure will be on forum

original-birdman:
> Reason for closure will be on forum

But it isn't. There are various comments indicating that no-one else has tried to code this, but feel that the current notifier system is perfect, see no reason to improve it and hence fell that a solution must be available without it.
Which isn't a good reason.

PR #154, opened 2017-04-02
Fix truncation of a utf8 filenames

original-birdman:


original-birdman:
See:
http://www.world-of-satellite.com/showthread.php?56078-File-timers-xml-damaged&p=445004&viewfull=1#post445004

ViX is reporting invalid `timer.xml` files because of the filename truncation (for length) being done in the middle of a multi-byte utf8 sequence, leaving invalid utf8.
This fix ensures the filename is valid utf8 (it discards any invalid utf8 parts).

The problem only exists for those using long filenames, and where those names, which are based on the programme description, contain multi-byte utf8 characters. So I can't test this in place, but I have tested that the line of code works by taking a string that ends in a multi-byte utf8 character and running the code on that string minus its last byte.

```
#!/usr/bin/python
# -*- coding: utf8 -*-

str = "Säsong låter för så får på rä"

i = len(str)
print "Orig len:", i

okstr = unicode(str[:-1], "utf8", 'ignore').encode('utf8', 'ignore')

print "Result (len=%d): %s" % (len(okstr), okstr)

```

PR #155, opened 2017-04-04
[ChannelSelectionContextMenu] fix shortcuts for nonexistent menu items

prl001:
ChannelSelectionContextMenu has a fixed ActionMap for its ChoiceList,
so that shortcut buttons (numbers, colours) are always active, even
if there is no menu item corresponding to the shortcut's action.

This can cause crashes and other inappropriate behaviour.

Now dynamically create the menu's ActionMap so that only the correct shortcut buttons are active.


prl001:
Added commit to fix button hint for "Add service to favourites".

PR #156, opened 2017-04-04
Provide backwards compatibility conversion for time zone settings

peterurbanec:
The new timezone code is not backwards compatible with existing settings.

This code will convert any setting that was valid with the old code to a matching new setting.

### DIFF
--- a/lib/python/Components/config.py
+++ b/lib/python/Components/config.py
@@ -1921,6 +1921,98 @@ class ConfigSubsection(object):
 	def dict(self):
 		return self.content.items
 
+# This converts old timezone settings string to the new two tier format.
+# The conversion table is based on the contents of timezone.xml just before it was removed from the source code.
+# If the old string is not found, the defaults are used instead.
+def convertOldTimezone(old_name):
+	old_zones = {
+		"(GMT-12:00) International Date Line West": ("Pacific", "Kwajalein"),
+		"(GMT-11:00) Midway Island, Samoa": ("Pacific", "Midway"),
+		"(GMT-10:00) Hawaii": ("Pacific", "Honolulu"),
+		"(GMT-09:00) Alaska": ("America", "Anchorage"),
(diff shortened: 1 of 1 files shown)


peterurbanec:
Updated with another commit that adds the extra timezones. I am intentionally trying to keep a number of versions that may have existed in the past so that it is possible to convert settings generated by older versions as well as recent versions. If there was a conflict, the more recent version is used.

andyblac:
sorry, but no, we longer use the xlm format.

IanSav:
But this code is not retaining the old XML format it simply offers forward based conversion of the old data, should it be encountered, in system backup images.

peterurbanec:
This code doesn't use the XML format.

The purpose is to provide backwards compatibility for user without silently resetting their timezone.

OpenViX changed the way timezones are handled in a way that is not backwards compatible. An update to a version that uses the new two-tier timezone code will silently reset the timezone for everyone to built in default (Europe/London), because the old value is not a valid setting in the new code base. As far as I am concerned, it's not acceptable to ship an update that will silently change the timezone.

This patch will try it's best to convert the old timezone setting to a matching pair of new settings. If successful, the previous user setting will be converted to the new format. If the conversion fails, the behaviour will be the same as current code and the default (Europe/London) will be used.

andyblac:
we are not adding backwards upgrade to new way, get used to it, just reset your TZ and be done with it.

IanSav:
In case you have missed it, the point PeterU is trying to convey is that any time in the future if a user installs an old backup (made prior to the new timezone system) the system will silently invoke the default timezone configuration.  The new timezone settings, as entered when the firmware upgrade is installed, are lost and the user never gets notified of the change / regression.

The submitted code simply tries to use the restored (old format) setting to try and calculate a better default in the new system.

If the rescue of the old data is what offends then perhaps the code can simply notify the user that the timezone settings have been lost during a backup restoration of an invalid timezone format.

PR #169, opened 2017-06-21
Ensure save current timeshift event doesn't sometimes save wrong file

prl001:
If InfoBarTimeshift.SaveTimeshift() is called without an explicit
timeshift file argument, and in the timeshift buffer directory
(/media/hdd/timeshift) the file timeshift.XXXXXX.sc is listed after
timeshift.XXXXXX (the .ts file) when os.listdir("/media/hdd/timeshift")
is called by SaveTimeshift() and timeshift.XXXXXX.sc has been updated
in the last 5 seconds, then timeshift.XXXXXX.sc will be used as the
saved timeshift buffer's .ts file, instead of timeshift.XXXXXX,
which is the actual TS file.

*** Replication steps ***

Change to a new channel, view ite timeshifted for a minute or two,
then change channels.

In the the "You seem to be in timeshift, Do you want to leave
timeshift ?" popup, select "Yes, but save timeshift as movie and
stop recording".

When timeshift has been saved check the size of the saved timeshift
file.  About half the time it will be small and not playable,
because the timeshift .sc file has been saved as the recording;'s
.ts instead of the timeshift TS file (no extension).

There is a more detailed bug report in the Beyonwiz bug tracker at
https://bitbucket.org/beyonwiz/easy-ui-4/issues/490/save-current-timeshift-event-sometimes

Fix:

Ensure that only the active timeshift.XXXXXX amd not the
timeshift.XXXXXX.ts file can be matched as savefilename in
InfoBarTimeshift.SaveTimeshift() when no timeshiftfile argument is
supplied.

This has been done by using some small functions to test whether a
filename has the correct timeshift.XXXXXX form in all tests in
InfoBarTimeshift that scan the files in the timeshift directory.

The fix was originally applied in the Beyonwiz easy-ui-4 code as a fix for
Beyonwiz Bug #375: Timeshift fails with misleading error message if
timeshift location contains directory named "timeshift".
https://bitbucket.org/beyonwiz/easy-ui-4/issues/375/timeshift-fails-with-misleading-error

While this fix does cleaner testing for timeshift.XXXXXX files, it
has more extensive changes than are strictly needed to fix the "save
timeshift" bug. If this causes difficilties with OpenViX upstream
merges, I can create a minimal fix that doesn't do the general
cleanup.

### DIFF
--- a/lib/python/Components/Timeshift.py
+++ b/lib/python/Components/Timeshift.py
@@ -55,6 +55,37 @@ from random import randint
 
 import os
 
+posixPortableFilenameChars = set("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-")
+
+# Is used for filename substrings, so does not enforce that
+# the name does not start with '-'
+
+def isPosixPortableFilename(fn):
+	return all(c in posixPortableFilenameChars for c in fn)
+
+# Tests on whether a filename appears to be generated from
+# mkstemp("timeshift.XXXXXX") as in lib/service/servicedvb.cpp.
+
+# mkstemp("timeshift.XXXXXX") with any suffix
+
+def isTimeshiftFilename(fn):
+	return len(fn) >= 16 and fn.startswith("timeshift.") and isPosixPortableFilename(fn[10:16])
+
+# mkstemp("timeshift.XXXXXX") with any suffix except ones ending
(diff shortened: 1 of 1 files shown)


Huevos:
Can you please the minimal bug fix... And maybe we can also push that to PLi if the bug also exists there too.

prl001:
OK, will try to do it tomorrow AU time.

PR #173, opened 2017-06-26
Fix crash on zap in RecordTimerEntry when timeshifted

prl001:
If a RecordTimerEntry needs to zap to a new service (zap timer,
zap+record timer or to avoid conflicts for a record timer) and
the user is currently timeshifted, the UI crashes.

The problem is due to what looks like an import of code from OpenPLi in
commit f90cfaa.

That import seems to have failed to adapt the code correctly for the
OpenViX and easy-ui-4 (Beyonwiz) images.

In RecordTimerEntry.openChoiceActionBeforeZap(),
self.InfoBarInstance.timeshiftActivated() is called, but
there is no such method in InfoBar (it would be expected to be in
the InfoBarTimeshift mixin if it existed).

In the zapAction() function defined inside
RecordTimerEntry.openChoiceActionBeforeZap(),
self.InfoBarInstance.saveTimeshiftFiles() is called,
which also doesn't exist in InfoBar.

Again in zapAction(), if choice is "save" or "save_movie"
self.InfoBarInstance.save_timeshift_file is set to True,
but it is never reset.

Again in zapAction, if choice is "save" or "save_movie",
self.InfoBarInstance.save_timeshift_in_movie_dir is set to True,
but the mechanism that this is intended to activate doesn't
exist in OpenViX or easy-ui-4.

Replication:

In live TV, initiate viewing timeshifted.

TIMER to open the timer list screen and add a new zap timer for a
few minutes ahead, and return to viewing timeshifted TV.

When the zap timer fires and the "You must switch to the service
... Timeshift is running. Select an action." popup appears, select
"Save timeshift and zap".

UI crashes.

The fix:

Removes the call to self.InfoBarInstance.timeshiftActivated()
and always inserts "Save timeshift and zap" in the first choice slot
if self.InfoBarInstance.save_timeshift_file is False (since
openChoiceActionBeforeZap() is only called when timeshift is being viewed
the test is redundant anyway).

Removes the "Save timeshift in movie dir and zap"/"save_movie" menu choice
and its associated code, since it does nothing in OpenViX or easy-ui-4.

Replaces the code in the "save" choice in zapAction() to use
self.InfoBarInstance.SaveTimeshift() instead of
self.InfoBarInstance.saveTimeshiftFiles().

Resets InfoBarTimeshift.save_timeshift_file in
self.InfoBarInstance.SaveTimeshift().

It's not entirely clear to me what the role of
InfoBarTimeshift.save_timeshift_file is intended to be, and I suspect
that it may not achieve what is intended.

### DIFF
--- a/RecordTimer.py
+++ b/RecordTimer.py
@@ -672,11 +672,7 @@ class RecordTimerEntry(timer.TimerEntry, object):
 			message += _("Timeshift is running. Select an action.\n")
 			choice = [(_("Zap"), "zap"), (_("Don't zap and disable timer"), "disable"), (_("Don't zap and remove timer"), "remove")]
 			if not self.InfoBarInstance.save_timeshift_file:
-				choice.insert(1, (_("Save timeshift in movie dir and zap"), "save_movie"))
-				if self.InfoBarInstance.timeshiftActivated():
-					choice.insert(0, (_("Save timeshift and zap"), "save"))
-				else:
-					choice.insert(1, (_("Save timeshift and zap"), "save"))
+				choice.insert(0, (_("Save timeshift and zap"), "save"))
 			else:
 				message += _("Reminder, you have chosen to save timeshift file.")
 			#if self.justplay or self.always_zap:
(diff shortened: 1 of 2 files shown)


AbuBaniaz:
I could not recreate this on a single tuner receiver (MB Micro premium) using Release 5.0 020. I did not even get a pop-up message. I had set the zap timer by the Graphical EPG instead of Timer option. (MB Micro drivers are poor, display sometimes does not show/disappear.) This is a test receiver so all the settings are default. 

Setting using Timer menu is a bit tricky, but results in no free tuner error, then a floating point exception and a restart without a crash log. Is that the same issue you are getting?

Main receiver is on Dev 5.0.021.003 kicks up an error about setting timer on remote receiver. I suspect this is a result of recent Webif changes. So could not test on that.

AbuBaniaz:
Please ignore above. Resolved issue with remote timer. I got the crash.
[Enigma2_crash_2017-06-26_11-32-00.zip](https://github.com/OpenViX/enigma2/files/1101798/Enigma2_crash_2017-06-26_11-32-00.zip)
```
<  1106.603> Traceback (most recent call last):
<  1106.603>   File "/usr/lib/enigma2/python/timer.py", line 239, in calcNextActivation
<  1106.603>   File "/usr/lib/enigma2/python/timer.py", line 309, in processActivation
<  1106.603>   File "/usr/lib/enigma2/python/RecordTimer.py", line 910, in doActivate
<  1106.603>   File "/usr/lib/enigma2/python/RecordTimer.py", line 486, in activate
<  1106.603>   File "/usr/lib/enigma2/python/RecordTimer.py", line 676, in openChoiceActionBeforeZap
<  1106.603> AttributeError: 'InfoBar' object has no attribute 'timeshiftActivated'
<  1106.603> [ePyObject] (CallObject(<bound method RecordTimer.calcNextActivation of <RecordTimer.RecordTimer instance at 0x928b9288>>,()) failed)
```

prl001:
Yes, that second one is the bug I saw and that this fix addresses. I was testing on a Beyonwiz T3 running OpenViX. I didn't see the floating point exception that you mentioned in the first comment. For my tests I was setting the timers using the T3's IR remote interface, not Open Webif.

PR #183, opened 2017-07-31
[ChannelSelection] Use fallback if no network time and not on DVB ser…

prl001:
…vice

If the initial live TV service at startup isn't a DVB service
(more accurately, isn't a service that supports the Time Of Day table)
and the system has no internet connection, the system time will
remain at the system startup default of 00:00:00 1 Jan 1970. An
example of such a non-broadcast service is the HDMI IN service on the
Beyonwiz T4.

This will mean that the displayed time on the PVR will be wrong,
any recordings that are due won't be made, an the default view of
the EPG will be for 1 Jan 1970.

Replication steps

If possible, select a live TV service that is not a DVB broadcast,
disconnect the PVR from the network and reboot.

If that isn't feasible, in the command-line interface:

init 4

Edit /etc/enigma2/settings so that config.tv.lastservice is a valid
service ref, but not a DVB or similar service (i.e. has on Time Of
Day table), e.g. 8192:0:1:0:0:0:0:0:0:0:. For that particular
example, it doesn't matter whether the device supports HDMI in.

Then:

halt

Then disconnect the PVR from the network and restart.

In either case, the time will start from 00:00:00 1 Jan 1970 when
the system restarts, and continue from that time setting until a
valid DVB channel is selected.

Fix:

If at startup config.tv.lastservice is not valid or, if it is valid,
but not a DVB service and the time is < 1 Jan 2004, find the first
valid DVB service in the current bouquet, or if that fails and
config.usage.multibouquet is set, then the first valid DVB service
in any other bouquet.

As well as fixing this bug, the fix will try to find a valid DVB
service if config.tv.lastservice is not valid.

The new methods ChannelSelection::findServiceBouquet() and
ChannelSelection::forceZap() are intended to be able to be called
to replace most of the code in RecordTimer::failureCB(). They may also
be able to be adapted to replace most of the code involved in
zapping by channel number (and the channel panic button action).

### DIFF
--- a/lib/python/Screens/ChannelSelection.py
+++ b/lib/python/Screens/ChannelSelection.py
@@ -17,7 +17,7 @@ from Components.Renderer.Picon import getPiconName
 from Screens.TimerEdit import TimerSanityConflict
 profile("ChannelSelection.py 1")
 from EpgSelection import EPGSelection
-from enigma import eActionMap, eServiceReference, eEPGCache, eServiceCenter, eRCInput, eTimer, ePoint, eDVBDB, iPlayableService, iServiceInformation, getPrevAsciiCode, eEnv, loadPNG
+from enigma import eActionMap, eServiceReference, eEPGCache, eServiceCenter, eRCInput, eTimer, ePoint, eDVBDB, iPlayableService, iServiceInformation, getPrevAsciiCode, eEnv, loadPNG, eDVBLocalTimeHandler
 from Components.config import config, configfile, ConfigSubsection, ConfigText, ConfigYesNo
 from Tools.NumericalTextInput import NumericalTextInput
 profile("ChannelSelection.py 2")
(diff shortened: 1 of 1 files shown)


AbuBaniaz:
Using a motorised dish, what would happen in following scenario:
Dish pointed at 28.2, I remove 28.2 services and then reboot.

AbuBaniaz:
Just tested this, receiver will not boot until I connect the network cable.

AbuBaniaz:
Please ignore previous comment.

IanSav:
There is at least one situation that this solution does not address.  What about the road warriors who use Enigma2 devices on the move (holiday caravaners, grey nomads etc).  There are times when no signal or Internet access are available but the units are still fully usable as a media player.

Perhaps an alternate solution could be to ask the user for a date and time at boot time if no other time source is currently available.

prl001:
There is one fairly frequent poster on the Beyonwiz forum who fits the "road warrior/grey nomad" description, and I don't recall him ever posting about problems with date setting.

That's only a sample of one, and it's not to say that there isn't such a problem, but it's not an issue I'd thought about when making these changes. I'm not sure how you'd go about testing for when to ask the question.

Anyway, it's an issue that can be addressed separately from this proposed fix.

prl001:
The use case that prompted this fix was on a Beyonwiz T4. In that PVR, HDMI IN can be set as the current service from bouquets.

If the user has HDMI IN as the current live service and shuts down, and the network isn't connected at restart for a recording, the system time remains in the early hours of 1 Jan 1970, and the recording will not be started, since it appears to be some 47 years in the future.

The problem was reported as a bug on the Beyonwiz forum.

IanSav:
I reported the nomad issue here as it was an OpenViX user who contacted me directly requesting a fix for the issue.

I believe the solution to this issue may be related to the currently proposed fix in that if the user can elect to be asked for the time then there is no need or desire to change the currently selected channel (as proposed in this fix).

Perhaps the option used can be a configuration item.  If time is not set at boot then (1) Select another channel, (2) Ask the user, or (3) Do nothing.

prl001:
This fix is mainly about unattended startup, for which "ask the user" isn't an option.

"Ask the user" could be triggered by a timer that fires, say, a minute after startup and checks whether the time has been set. It still wouldn't be any use for unattended startup, but it would allow manual time entry.

Provided there is a service available that carries the Time Of Day table, the user can always set the time semi-manually by zapping to that service.

If the box is simply being used as a media player, it doesn't matter much to the system that the time is wrong.

This fix is aimed specifically at a scenario where the lack of network connections causes unattended recordings to fail.

IanSav:
Your point about this fix being targeted to rescue potentially failed recordings is very well made.

The request made to me was to address incorrect times being displayed on an otherwise functioning media player.

Hopefully a fix for the generic time issue will spring to mind and won't require a hack or much effort.

prl001:
I suggest that the issue of timesetting when there is neither network nor a viable way of getting time from signal Time Of Day tables be discussed on either or both of the OpenViX or Beyonwiz forums to get a workable scheme for when to ask the user to set the time.

I don't think that a simple Use another channel/Ask user/Do nothing setting will adequately cover both attended and unattended startups.

If there's a bug report or enhancement request added to the Beyonwiz issue tracker, I'll have a look at implementing it.

My inclination is to have a "when all else fails, ask user" with a reasonably long timeout (perhaps 3 min) that falls back to "do nothing".

Any such time setting will need to trigger the eDVBLocalTimeHandler::eDVBLocalTimeHandler signal to get all the stuff that needs to be notified of the time being set is fired off.

RobvanderDoes:
I'm one of the 'grey nomads' who runs into this issue quite often. Showing the wrong time is of course not a serious issue, but it is annoying as one has to remember that the date/time shown by the box is wrong.

> My inclination is to have a "when all else fails, ask user" with a reasonably long timeout (perhaps 3 min) that falls back to "do nothing".

Speaking for myself: I wouldn't even need that. A simple extra line in the time settings to allow manual time setting (the same as smart-TV's and Android devices have) would be sufficient.

prl001:
I would have thought it cleaner to set the time from the network or the broadcast if possible, and ask the user if it can't be set from either. I'd have no problem with a setting to always ask if that was thought necessary.

The time setting popup could show the correct current time if available, and may be able to get a reasonable first guess at the time from something like the last modified time of epg.dat.

I've already said that I'd be happy to try to implement/fix a reasonably well thought issue ticket. I just think its a separate issue from the one that this fix is trying to address.

And I've got fix for another problem related to clock-setting in the pipeline.

prl001:
Just for the information of people reviewing this pull request: I don't intend to make any further commits on this request. Any further changes (like a possible implementation of user time entry) will be made in separate pull requests.

PR #187, opened 2017-08-27
TryQuitMainloop can return, so clean-up properly if it does. Factor o…

original-birdman:
…ut common code to new NSCommon class to achieve this.


original-birdman:
There were several (9) classes with many (12) mostly-identical functions. These have been moved to a single NSCommon class.
The fix (to tidy up if `self.session.open(TryQuitMainloop, 2)` returns) was then added here, in one place.

PR #191, opened 2017-09-14
Update now/next info from EPG as soon as possible

prl001:
If MENU>Setup>EPG>Settings>Show EIT now/next in infobar is disabled and
there is no network connection at boot time, then there is no
now/next info in the infobar for 60 seconds after startup completes
or until the channel is changed, whichever is first.

Similarly, if Show EIT now/next in infobar disabled while using the
EIT EPG, and you change channels to a service that doesn't yet have
current EPG data. There's a 60-second wait before the now/next
information becomes available, even though the required EPG data
is loaded well before that.

The latter issue affects Australian users where each broadcaster's
mux carries only the EPG for that broadcaster. Other broadcast
areas may not be similarly affected.

Replication steps - boottime issue

Disable MENU>Setup>TV>EPG>Show EIT now/next in infobar.

Disconnect the PVR from the network (e.g. unplug Ethernet and any WiFi.

Reboot (GUI restart is not sufficient).

When the startup has completed, no now/next information is displayed
in the infobar. Redisplay the infobar at short intervals - 60 sec
after the startup completed, the now/next information becomes
available.

Reboot.

When the startup has completed, no now/next information is displayed
in the infobar. Before 60 seconds has elapsed and after the time
is correctly set from the broadcast, change channels. The now/next
information for the new channel is displayed immediately.

Replication steps - channel change issue

In the command-line interface to the PVR, clear the saved EPG data using:

init 4
rm /etc/enigma2/epg.dat
init 3

When the GUI has restarted and the broadcast EPG for the current
channel has been loaded, use FAV to select a channel where the
broadcast EPG has not yet been loaded.

No now/next information is displayed in the infobar for the newly
selected channel. Redisplay the infobar at short intervals - 60 sec
after the startup completed, the now/next information becomes
available.

Fix

If MENU>Setup>EPG>Settings>Show EIT now/next in infobar is disabled
and the system time was not set at startup or there is no current
EPG information for the channel, connect
eDVBServicePlay::updateEpgCacheNowNext() to a new
eEPGCache::cacheUpdated() signal so that the now/next information
can be extracted from the EPG cache as soon as the information
is available in the EPG cache.

Disconnect the signal connection as soon as valid now/next data
has become available.

This fixes both issues outlined above.


Huevos:
This request hasn't been forgotten.

prl001:
Hi, Huevos. I haven't forgotten about it either. It's just as well that it hasn't been merged, because I've since worked out a cleaner way of doing it. I'm having problems building the OpenViX build environment at the moment, but once that's done, I'll re-submit in a new branch and pull request.

While this fixes a small, uncommonly encountered bug, I would also like to have the capability that the fix adds - the ability to be notified on EPG updates - to use for other planned improvements. Especially updating the details of a repeat timer as soon as they become available, instead of just before its next run. I'll make a note of this when I re-submit it.

RobvanderDoes:
> Especially updating the details of a repeat timer as soon as they become available, instead of just before its next run. 

Can you explain the advantages (and disadvantages?) of that please? I can think of some, but apparently you have thought this over.

prl001:
> > Especially updating the details of a repeat timer as soon as they become available, instead of just before its next run.
> 
> Can you explain the advantages (and disadvantages?) of that please? I can think of some, but apparently you have thought this over.

Sure. The main problem is that the timer EIT id of repeat timers is never updated at all, even when the timer description is fetched and put into the recording metadata just before recording.

This means that once a repeat timer has been run for the first time, it cannot be accessed via "Change timer" on the green button in the EPG, because the green button switches from "Add timer" to "Change timer" if the EIT id of the EPG event in focus matches the EIT id of a timer.

This bug is flagged in [issue #174](https://bitbucket.org/beyonwiz/easy-ui-4/issues/174/graphical-epg-highlights-active-timers-but) in the Beyonwiz issue tracker (and was logged back in 2014). Note that step 2 in the Reproduction steps is not necessary, the test is only on the EIT id and ignores the event times - there's a comment in the issue to that effect.

A secondary issue is that while an up-to-date description is (or optionally can be) written into the recording .ts.meta file when the timer records, the timer description will remain set to its original setting, rather than showing the description for the event that the timer will record next.

For timers whose repetition time is less than the EPG time span, the update can be done when the timer completes and updates its time to the next event to record. However, for weekly timers, in Australia, the next event is typically not available in the EPG to update the EIT id and description.

Having an event generated on EPG update completion would allow such timers to be updated to the correct EIT id and description when the EPG updates and adds the next event for timers that couldn't be updated correctly.

The down side is that it is not possible to unambiguously convert a timer start/end time to an event, and so to an EIT id or description, because padding time is not described in the timer.

This is a particular problem in Australia, where the broadcasters (especially, but not exclusively, the commercial broadcasters) do not broadcast their programs accurately to the published schedule (as a normal practice, not just when caused by a big breaking news story, or sport running over (or under) time). I have pre-padding of 5 minutes and post-padding of 25 minutes and that is fairly typical for Australian PVR users.

That means that if a program is shorter than the padding time, trying to get the EIT id and description using a probe using

eEPGCache.getInstance().lookupEventTime(rec_ref, (rec_start+rec_end)/2)

may not match the correct program. This is particularly the case for children's programming, where events are often 30 min or less.

My understanding is that, in general, broadcast schedules are better adhered to by European broadcasters, and that problem may not be so important for European users.

The "timer times to event" mapping ambiguity could be much improved by saving the padding times explicitly in the timer, but that's an argument for another day :)

I think that covers everything. My apologies for the length. If anything is unclear, please ask.

I intend to submit both this bug fix and its use to do those timer updates to the Beyonwiz repository anyway.

RobvanderDoes:
Thanks for your exhaustive explanation 👍 

> This means that once a repeat timer has been run for the first time, it cannot be accessed via "Change timer" on the green button in the EPG,

I can confirm this issue, but never understood the underlying cause. This happens quite often to me.

> My understanding is that, in general, broadcast schedules are better adhered to by European broadcasters, and that problem may not be so important for European users.

This can't be said as general statement. 
e.g. for UK TV this is indeed true, but for Dutch & Belgium TV I need to use the 5 mins before and 30 mins after to correctly make recordings (and even then I sometimes miss the last few minutes).
And indeed kids programs are often (very) much shorter, starting from around 3 mins.

RobvanderDoes:
BTW, regarding

> The "timer times to event" mapping ambiguity could be much improved by saving the padding times explicitly in the timer, but that's an argument for another day :)

won't it be easier to 'fix' this first?

And related to that: there's now one default for pre- and post- padding time. In my case I don t need any padding time for recordings from UK-TV, while I do need the values I mentioned for Dutch & Belgium TV. Would a default per-service (or per group of services) be an option?

prl001:
> Thanks for your exhaustive explanation 👍

Thanks.

>>    My understanding is that, in general, broadcast schedules are better adhered to by European broadcasters, and that problem may not be so important for European users.
>
> This can't be said as general statement.
e.g. for UK TV this is indeed true, but for Dutch & Belgium TV I need to use the 5 mins before and 30 mins after to correctly make recordings (and even then I sometimes miss the last few minutes).
And indeed kids programs are often (very) much shorter, starting from around 3 mins.

I wasn't aware of that. I'd thought that European broadcast regulation on scheduling was a good deal stronger generally than in Australia (where it's pretty lax). I guess, even with the EU, it's still foolish to expect too much uniformity of regulation in Europe.

prl001:
>> The "timer times to event" mapping ambiguity could be much improved by saving the padding times explicitly in the timer, but that's an argument for another day :)
>
> won't it be easier to 'fix' this first?

"Easy" isn't something I'd attach to this. There have been a couple of discussions about how it might be done on the Beyonwiz forum. I can dig out links to them if you think they may be interesting.

The big problem is getting any such changes done so that they integrate smoothly into plugins, especially ones that are quite sensitive to any changes in the meaning of timer begin & end, like AutoTimer.

I have considered doing it by re-implementing TimerEntry.begin and TimerEntry.end as properties, and having the nominal event start/end and timer padding stored in the timer. In particular, though, this would still need AutoTimer to supply the explicit padding values for timers it generates.

If it were easy, I might have already done it.

>And related to that: there's now one default for pre- and post- padding time. In my case I don t need any padding time for recordings from UK-TV, while I do need the values I mentioned for Dutch & Belgium TV. Would a default per-service (or per group of services) be an option?

That's certainly a possibility. There's a frequent poster on the Beyonwiz forum who handcrafts timer padding to suit the habits of particular broadcasters, generally less padding for the national broadcasters ABC and SBS, and more for the commercial broadcasters.

The lamedb provider name would probably be suitable to identify the broadcaster in FTA terrestrial broadcasts in Australia. Satellite, I'm sure, would add some complexity.

RobvanderDoes:
> >won't it be easier to 'fix' this first?

> "Easy" isn't something I'd attach to this. 

Sorry; that must be my non-native English; what I meant to say is "wouldn't it be better to fix this first".

Anyway: for testing and hence 'easy' committing any pullrequest I strongly suggest to embed only one fix/enhancement per request, if feasible.

prl001:
> Sorry; that must be my non-native English; what I meant to say is "wouldn't it be better to fix this first".

No need to apologise. And yes, it would be better to fix it first. I'm just not sure how best to do it.

> Anyway: for testing and hence 'easy' committing any pullrequest I strongly suggest to embed only one fix/enhancement per request, if feasible.

Absolutely. My intention is to revise this fix so that it has a cleaner implementation and issue a separate pull request, and to do any changes that use the "EPG updated" event in other separate pull requests (that code is also not yet written). That's how I'll do it on the Beyonwiz repository, anyway.

prl001:
A bit off-topic: what branch of https://github.com/oe-alliance/build-enviroment should I be using for OpenViX builds?

Huevos:
At the moment branch 4.1.

prl001:
Thanks, Huevos.

PR #193, opened 2017-09-28
If next timer is before now+MaxWaitTime then remember and use it!

original-birdman:
Without this a one-off DeepStandby PowerTimer will, more often than not, fail to work.


original-birdman:
See:
http://www.world-of-satellite.com/showthread.php?57337-Sleep-Timer-Function-on-Vix-images&p=457863&viewfull=1#post457863
for the report of the problem leading to this PR.

PR #194, opened 2017-09-28
Set afterevent to NONE for a Zap-only timer to prevent a possible shu…

original-birdman:
…tdown attempt at zap-time.

### DIFF
--- a/lib/python/Screens/TimerEntry.py
+++ b/lib/python/Screens/TimerEntry.py
@@ -450,6 +450,13 @@ class TimerEntry(Screen, ConfigListScreen):
 			"standby": AFTEREVENT.STANDBY,
 			"auto": AFTEREVENT.AUTO
 			}[self.timerentry_afterevent.value]
+# There is no point doing anything after a Zap-only timer!
+# For a start, you can't actually configure anything in the menu, but
+# leaving it as AUTO means that the code may try to shutdown at Zap time
+# if the Zap timer woke the box up.
+#
+		if self.timer.justplay:
+			self.timer.afterEvent = AFTEREVENT.NONE
 		self.timer.descramble = {
 			"normal": True,
 			"descrambled+ecm": True,


IanSav:
Is this a wise change?

I think it would be better to make ZAP timers more like real timers in that they have a nominated duration.

This would mean that ZAP timers will correctly show in the EPG.  It would also allow users to nominate to have a usable after event action.  For example, a user may want to have a ZAP timer start the box display a specific show, like a news broadcast, and then shut down after the timer.  This could be used as a morning wake up alarm etc.

original-birdman:
> Is this a wise change?

Yes. The only point of a Zap timer is to set the channel. As such it has no duration.
Since they have no duration (once the channel has changed the timer is complete) they can't show in an EPG. There is no mechanism for "hogging" a tuner/transponder apart from recording from it.

You can a achieve your morning wake call with a Zap timer + PowerTimer to shutdown.
However, if it's woken you up you should be able to shut it down manually....

IanSav:
A ZAP timer *could* have a duration.  :)  I have seen PVRs where ZAP timers are shown like other timers and it can be helpful.

The ZAP timer doesn't need to hog a real tuner, it could simply change the current service and then release the tuner.  If a resource must be allocated against the tuner then perhaps a dummy tuner can be created for this purpose.

In the 'old' days I would have asked for the tuner to be locked like for a real RECORD timer as a common usage was to set a ZAP (or VIEW as we used to call them) timer to create a video output feed to be used by an external device (like a VCR).  In that case you didn't want the current service to change or be changeable until the timer completed.

Needing to set a ZAP timers and then co-ordinate it with a POWER timer is fiddly.  Allowing the ZAP timer to have a nominal duration makes the process simple for everyone.

Perhaps a new timer type called a VIEW timer can be created to address the mode of use I suggested above.  Alternatively the current ZAP timer could be implemented as described above via a configurable option.

original-birdman:
> The ZAP timer doesn't need to hog a real tuner, it could simply change the current service and then release the tuner.

That's what it does at the moment....

This fix is to prevent this problem:
http://www.world-of-satellite.com/showthread.php?57904-Unwanted-message

IanSav:
I accept that this change addresses the problem as described in the forum.

I think that the underlying issue is the poor understanding of what an Enigma2  ZAP timer is all about.  In reality the ZAP timer shouldn't be in the same class as a RECORD timer.  It would be better suited as a POWER timer type.  (At the nominated time change the current service.  If the unit is off then power on to achieve the change.)  It is an instantaneous transition type event.  RECORD timers are a duration based event.  (This is how the VIEW timers worked (as duration type timers) on the other box I mentioned.)

Huevos:
Zap timer do not have duration. Their sole purpose is to zap. They are not a workaround solution for other problems.

IanSav:
My thoughts are heavily influenced by past experiences on other products.

original-birdman:
I can see the possibility for a VIEW recording timer. You could set one for something you knew you were going to watch live, and hope to push any clashing timers onto other time slots....

However, that would be _much_ simpler to code as a Recording timer with no recording file (since everything else is then in place for clash detection) rather than a Zap timer with a duration.

original-birdman:
>  In reality the ZAP timer shouldn't be in the same class as a RECORD timer.

Yes, it should. It manipulates tuners. Power timers only change the power state of the box.

AbuBaniaz:
> There is no mechanism for "hogging" a tuner/transponder apart from recording from it.

Yes there isn't directly. But fake recording as used by EPG refresh does so.  I'd like that to be added into timers so we can dispense with the need for the zapping plugin.

original-birdman:
So you'd like a VIEW recording option to be added?  That might be fairly simple, at least for the timer coding - the menu to set it might provide more of a challenge...

IanSav:
In for a penny, in for a pound:

I believe that from a User Interface and User Experience point of view the current mechanisms for setting timers is not particularly optimal or user friendly.

I believe that a single menu item should be dedicated to timers of all types.  (Except AutoTimers and Cron Timers that are external, optional, plugins.)  All the timers should be found under the MENU > Timers area and can also be triggered by pressing the TIMER button, if one exists, on the remote control.

The combined list of timers could include:

- Record (Wake system from Deep Standby or Standby, start normal background recording and at the end of the duration carry out the After Event.)
- Zap and Record (Wake system from Deep Standby or Standby, change LIVE TV service and start foreground recording and at the end of the duration carry out the After Event.)
- View (Formally Wakeup - Wake system from Deep Standby or Standby, change the foreground /LIVE TV service without recording it and then at the end of the duration carry out the After Event.)
- Zap (Wake system from Deep Standby or Standby and change the foreground / LIVE TV service without recording it.  This timer has no duration or After Event.)
- Wake to Standby (Wake system from Deep Standby to Standby.  Foreground / LIVE TV service is unchanged  then at the end of the duration carry out the After Event. Used to allow background tasks to complete or to allow remote access at nominated times.)
- Wake (Wake system from Deep Standby or Standby.  Foreground / LIVE TV service is unchanged  then at the end of the duration carry out the After Event. Used to allow background tasks to complete or to allow remote access at nominated times.)
- The Wake to Standby and Wake timers can be combined / simplified as:
Wake (Wake the system to the nominated operating level - Standby or Running.  Foreground / LIVE TV service is unchanged  then at the end of the duration carry out the After Event. Used to allow background tasks to complete or to allow remote access at nominated times.)
- Auto Standby (From the current state wait for the duration to complete after the last user initiated command and then go to Standby.)
- Auto Deep Standby (From the current state wait for the duration to complete after the last user initiated command and then go to Deep Standby.)
- The Auto Standby and Auto Deep Standby timers can be combined / simplified as:
Sleep (From the current state wait for the duration to complete after the last user initiated command and then carry out the After Event - Standby or Deep Standby.)
- Go to Standby
- Go to DeepStandby
- Reboot system (At the nominated date and time the system will be rebooted and then carry out the After Event - Go to Standby, Go to Deep Standby or Do nothing.)
- Restart GUI (At the nominated date and time the GUI will be restarted and then carry out the After Event - Go to Standby, Go to Deep Standby or Do nothing.)

If the optimisations are used then the Timer menu would consist of:

- Record
- Zap and Record
- View
- Zap
- Wake
- Sleep
- Go to Standby
- Go to Deep Standby
- Reboot system
- Restart GUI

I know this is a bit radical but I thought the ideas should be expressed to encourage thought and discussion with a possible view to future adoption.

PR #215, opened 2018-02-05
Allow any serviceref in EPGList::getIndexFromService(), not just thos…

original-birdman:
…e starting "1:" - it's either there or not.

In EPGList::getSelectionPosition(), use self.l.getCurrentSelectionIndex() for all cases (it is set...) and simplify code.

### DIFF
--- a/lib/python/Components/EpgList.py
+++ b/lib/python/Components/EpgList.py
@@ -333,14 +333,10 @@ class EPGList(GUIComponent):
 	def getIndexFromService(self, serviceref):
 		if serviceref is not None:
 			for x in range(len(self.list)):
-				if str(self.list[x][0]).startswith('1:'): # check for Graphical EPG
-					if CompareWithAlternatives(self.list[x][0], serviceref.toString()):
-						return x
-				elif str(self.list[x][1]).startswith('1:'): # check for Multi EPG
-					if CompareWithAlternatives(self.list[x][1], serviceref.toString()):
-						return x
-				else:
-					return None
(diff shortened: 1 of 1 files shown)


Huevos:
Gordon, can you explain to a layman what this does and how to test, please:

original-birdman:
There are two parts to this.

`getIndexFromService()` is looking through the list of services in the current (displayed) bouquet to find which one is the current service. It's just looking for the index number (0 to n-1).
For some reason it was testing that the service ref, as a string, started with "1:", but that makes no sense, as an IPTV service will start with "4097:", but you'll still want to find it. In fact, you'll want to find any entry which is actually there, so testing _any_ sub-string makes no sense. So I removed the erroneous test.

In the "if" blocks, the code is just looking for where on the screen the current selection is so that it can pop-up the Add Time/Add AutoTimer overlay. For two of the cases it was calling `getIndexFromService()` but given that getCurrentSelectionIndex() is always valid it is simpler just to call that. This then makes all 5 cases use the same call, so the call can be brought out of each case. The cases then are just used to set the step variable to find out the relative offset on the current display.

The testing has been done - see:
http://www.world-of-satellite.com/showthread.php?58891-Crash-when-recording-IPTV-from-G-EPG&p=465859&viewfull=1#post465859

original-birdman:
Oh, I forgot. In the original `getIndexFromService()` code, if you were searching for a service that started with "1:" but a preceding entry did _not_ then it wouldn't find the one you were looking for even if it were actually there. It would return `None` and the calling code in the "if" blocks would crash.

PR #217, opened 2018-02-13
Send the filetype along when getting Exif info, otherwise the code al…

original-birdman:
…ways thinks it has a JPEG.


original-birdman:
Since the ViX version has had various commits reverted and PLI ones applied, this PR is no longer needed (PLi already had this in).

PR #220, opened 2018-02-24
[NetworkSetup.py] Correctly enable VirtualKeyboard

IanSav:
This fix enables the VirtualKeyboard to be displayed when the TEXT button is enabled (on text entry items in the configuration list).


IanSav:
I omitted to say that this applies to uShare.

RobvanderDoes:
It applies to the screen 'NetworkuShareSetup' that is used by both the context menu's of 'uShare' & 'MiniDNLA'

PR #221, opened 2018-03-01
Remove use of non-existent self variable.

original-birdman:
This produced errors:
<    25.307> [InitNimManager] tunerTypeChanged error:  global name 'self' is not defined

[you'll now get "tunerTypeChanged to 'DVB-T2' failed (BUSY)" instead, if "another linked frontend is in use.. so allocateFrontendByIndex not possible!"
but at least not an error about a non-existent variable]

### DIFF
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -15,6 +15,8 @@ from Tools.BoundFunction import boundFunction
 from Components.About import about
 from config import config, ConfigSubsection, ConfigSelection, ConfigFloat, ConfigSatlist, ConfigYesNo, ConfigInteger, ConfigSubList, ConfigNothing, ConfigSubDict, ConfigOnOff, ConfigDateTime, ConfigText
 
+import NavigationInstance
+
 config.unicable = ConfigSubsection()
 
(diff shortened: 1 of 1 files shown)


Huevos:
How do you provoke this error. Please provide a step by step.

original-birdman:
> The error message you posted is just one line.

The error message _is_ just one line.

> How do you provoke this error.

All you need to do is boot a system with multiple DVB-C/DVB-T2 tuners that are configured to run as DVB-T2.
Happens on both my Xtrend et8000 and ccs's et10k:
   http://www.world-of-satellite.com/showthread.php?57967-Xtrend-Et10k-Timer-Recording-Didn%92t-Records-Properly&p=467176&viewfull=1#post467176
(It doesn't happen on my MBTwin, as the DVB-T2 tuners there are only DVB-T2 and hence never need to be changed. The same would apply to any DVB-S(2) tuner systems.). It won't happen on a system that doesn't have to change the type at boot-up.

If you look at the code you can see that there is no "self" defined in the function, so it's clearly wrong.

Huevos:
Thanks. This code came from PLi, so I am going to start a thread on their forum directing them to this commit.

PR #222, opened 2018-03-01
Add a justLoad option to loadTimer() (passed on to record()) to disab…

original-birdman:
…le conflict testing on loading. Needed when loading an initial backup, as the tuner configuration isn't necessarily correct then.

### DIFF
--- a/RecordTimer.py
+++ b/RecordTimer.py
@@ -997,7 +997,9 @@ class RecordTimer(timer.Timer):
 				return True
 		return False
 
-	def loadTimer(self):
+# justLoad is passed on to record()
+#
+	def loadTimer(self, justLoad=False):
 		try:
 			file = open(self.Filename, 'r')
 			doc = xml.etree.cElementTree.parse(file)
@@ -1025,7 +1027,7 @@ class RecordTimer(timer.Timer):
 		timer_text = ""
 		for timer in root.findall("timer"):
 			newTimer = createTimer(timer)
-			conflict_list = self.record(newTimer, ignoreTSC=True, dosave=False, loadtimer=True)
+			conflict_list = self.record(newTimer, ignoreTSC=True, dosave=False, loadtimer=True, justLoad=justLoad)
 			if conflict_list:
 				checkit = True
 				if newTimer in conflict_list:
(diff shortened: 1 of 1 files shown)


original-birdman:
Note that this needs:
   https://github.com/OpenViX/vix-core/pull/16
to actually _work_, but since it only adds an optional parameter call it doesn't fail if that is missing.

original-birdman:
Without this fix an initial restore (_e.g._ after a flash) will likely disable the second recording timer, at least on a DVB-T2 system where the tuners default to DVB-C.
See http://www.world-of-satellite.com/showthread.php?57967-Xtrend-Et10k-Timer-Recording-Didn%92t-Records-Properly&p=464892&viewfull=1#post464892 _et. seq._

original-birdman:
What else do I need to do to get acceptance, or some comment, on these two PRs?

Huevos:
Gordon, why is the tuner config not correct at load time? Is it due to the NimManager bug? Or is it a load order thing?

original-birdman:
It's certainly the case that the Record timers are loaded before the config file is loaded. 
http://www.world-of-satellite.com/showthread.php?57967-Xtrend-Et10k-Timer-Recording-Didn%92t-Records-Properly&p=467105&viewfull=1#post467105
http://www.world-of-satellite.com/showthread.php?57967-Xtrend-Et10k-Timer-Recording-Didn%92t-Records-Properly&p=467150&viewfull=1#post467150
But changing this order did not remove the problem:
http://www.world-of-satellite.com/showthread.php?57967-Xtrend-Et10k-Timer-Recording-Didn%92t-Records-Properly&p=467154&viewfull=1#post467154
So I tried re-initializing the `nimmanager` too - still no good:
http://www.world-of-satellite.com/showthread.php?57967-Xtrend-Et10k-Timer-Recording-Didn%92t-Records-Properly&p=467159&viewfull=1#post467159

However, even if all this had worked there is _another_ issue, which no amount of config loading/initializing re-ordering could change - that issue is that if you have a USB tuner it won't be visible _at all_ until after the plugins are restored (as that happens at a later stage).
http://www.world-of-satellite.com/showthread.php?57967-Xtrend-Et10k-Timer-Recording-Didn%92t-Records-Properly&p=467164&viewfull=1#post467164

So the only solution (apart from, possibly, _completely_ re-ordering the whole Restore Wizard and Backup manage code) is to ensure that all of the timers get restored regardless of the apparent tuner setting, bearing in mind that they have come from a restore from presumably the same system with the same real tuner config, and  they were OK at the time the backup was made.

SimonCapewell:
What would happen if the tuners had been reconfigured in the meantime? Let's say I swap one of them from a DVB-T2 to a DVB-S. Would the restore ignore any timers that don't make sense for the specific tuner? If so, does the restore warn about this?

AbuBaniaz:
The restoration and validation of timers are different items. Lets not fall into the trap of combining the two

original-birdman:
> What would happen if the tuners had been reconfigured in the meantime? 

Then this would be spotted the next time that any timer were added.
But if you have a USB tuner then the configuration will _always_ be wrong when you load the recording timers and this code fixes that issue.

original-birdman:
> The restoration and validation of timers are different items. Lets not fall into the trap of combining the two

They are not. Timers _are_ validated as they are restored, as each is added to the internal timer list, which, without this fix, runs TimerSanityCheck on each addition. And if you have a USB tuner (or a DVB-C/DVB-T2 one set to DVB-T2) your tuner config _will_ be wrong, so the validation _will not_ be valid.

Huevos:
Gordon I am going to merge this... but don't forget any backup can be restored on to any receiver.

I often share backups between the SF4008 and the OS Mega.

original-birdman:
> I often share backups between the SF4008 and the OS Mega.

In which case. unless you record the same things on the two boxes all of the time, the timers that are restored will be wrong anyway.

PR #225, opened 2018-03-10
[Timezones] Unnecessary instances of AutoTimer and AutoPoll created

prl001:
Change Components.Timezones to operate on
Plugins.Extensions.AutoTimer.plugin.autotimer and autopoller instead
of new instances of AutoTimer and AutoPoller.

Make the poll delay after a TZ change symbolic.

The attempts in the original code to stop the AutoPoller while
running AutoTimer.parseEPG() fail because they only stop the new
instance of AutoPoller, not the one in the plugin.


ViX-Sicilian:

PR #226, opened 2018-03-11
[Setup.py] Improve getSetupTitle

IanSav:
This change eliminates the forced crash if a setup menu id is not found in the setup.xml file.  This sort of error should be identified during development and not reach production code.  Forcing a crash is unnecessary and anti-social.  The crash could have nasty side effects like stopping running recordings.  Instead of the crash an error condition is logged and the user is provided with a dummy menu item indicating that this menu item is unavailable.  The error contains the ID tag so that the issue can be reported to developers for correction.

Blank "title=" attributes now have a default value of "Settings" to stop blank titles being fed into the translation system.

To test the change edit setup.xml and change the "key=" attribute for an item in a Setup based menu.  Before the code change loading the menu containing this menu item will force a crash of the GUI.  After the code change an error will be logged and the menu will be displayed but the menu item that can't be found will be shown as an error message.  (It is not fatal to select this menu item.  If selected a blank Setup screen will be shown.)

### DIFF
--- a/lib/python/Screens/Setup.py
+++ b/lib/python/Screens/Setup.py
@@ -33,12 +33,6 @@ def getConfigMenuItem(configElement):
 			return _(item.attrib["text"]), eval(configElement)
 	return "", None
 
-class SetupError(Exception):
-	def __init__(self, message):
-		self.msg = message
-
-	def __str__(self):
-		return self.msg
 
 class SetupSummary(Screen):
 	def __init__(self, session, parent):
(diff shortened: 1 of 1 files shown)


IanSav:
Unfortunately OpenViX and OpenPLi do not share comparable code.  There is a comparable change filed with OpenPLi.

IanSav:
What is the problem with this pull request?

Huevos:
It removes the exception. I thought we had already discussed this. The exception will not be reached in production code. Is there a valid reason not to raise the exception if the ID is missing? Please don't say graceful handling. Getting through to a blank screen is not graceful handling anyway.

IanSav:
The exception may not be intentionally reached but it is not impossible.  If the setup.xml file or the menu file that has the key references are damaged in any way the system can spontaneously crash for no reasonably acceptable reason.  The submitted code change provides logged and user feedback of the error situation without forcing a crash.

My point about the blank Setup screen is that if the user ignores the reasonably obvious error message and tries to load the menu anyway the system remains totally stable and simply shows an empty Setup screen.  Potentially aborting their recordings for a totally recoverable error should always be unacceptable.

Did anyone else have an opportunity to review the submission?  Is anyone else permitted to have an opinion on this request?  Did you even test the code to see how the new code works?

andyblac:
if i'm being honest i don't understand what your trying to achieve ?, you do release that you can use custom setup.xml anywhere in the image, by not touching the system one right ?

IanSav:
This is not a matter of writing new code for a new use for setup.xml.  I am trying to improve some existing code that is poorly written:
- The current code, if triggered, terminates all other Enigma2 tasks including any active recordings. 
- There is no need to crash the UI for a non fatal error.  In this case it is enough to display an error message and keep going.  (While crashing the UI may suit some developers while in development mode this is production code.)
- If there is no "titleshort" attribute, or it is an empty string, and the "title" attribute is also missing or blank then the code tries to translate an empty string.  It is inappropriate to translate empty strings.  (This generates a translation status message and not an empty string.)

This is the first of a number of improvements I have lined up for Setup.py

PR #228, opened 2018-03-14
Fix pixmap renderer not cleared on CHANGED_CLEAR

prl001:
Clear the pixmap displayed in Components.Renderer.Pixmap when changed()
receives a CHANGED_CLEAR.

CHANGED_CLEAR is issued upstream in sources like Event and EventInfo
when no event can be found, and so this change clears the displayed
pixmap when no pixmap can be found because no upstream event has
been found.

Previously CHANGED_CLEAR was ignored.


ViX-Sicilian:

PR #229, opened 2018-03-14
[ValueToPixmap] Fix potential crash

prl001:
Components.Converter.ValueToPixmap.getPixmap() tests whether
self.source is valid, but if it's not, it continues on to use
an undefined variable (val).

Change the code so that None is returned if self.source is not valid.


ViX-Sicilian:

PR #231, opened 2018-03-14
Set self.MountPath in calculateFilename if it is not set.

original-birdman:


original-birdman:
Vix has added a `freespace()` call, which sets self.MountPath, which `calculateFilename()` then uses. However, external plugins (such as the Series Plugin in `enigma2-plugins`) won't make the `freespace()`call as it is not in all distros.
There is no point in calling `freespace()` itself as the caller won't be able to handle anything it does anyway.

Huevos:
Can the plugin not be made to work in all distros, rather than hijacking our enigma?

original-birdman:
It's Vix which has introduced the new call, so it has to be Vix that handles it.

original-birdman:
Skipping what?
If we skip the `freespace()` call we end up failing recordings set to record on a non-existent file-system when we do have space in the default location (this is what `freespace()` does).

PR #240, opened 2018-03-21
Ensure that the timer repeated field is cleared if we have an autosle…

original-birdman:
…eprepeat of "repeated".


original-birdman:
Without this the presence of the repeated field for the timer means that it is used to set the next timer time, rather then the autosleepdelay field.
The `PowerTimerEntry,py` fix ensures that this happens when a timer is edited (its type codul be changed...)
The `PowerTimer.py` fix ensures that this happens for timers that are loaded from file, which will fix any existing timers that are "wrong".

PR #241, opened 2018-03-21
Check that the item to be changed exists in the four functions to in/…

original-birdman:
…decrement times to avoid a crash.


original-birdman:
This also correct a typo in some variables (`timerrntry_` should be `timerentry_`) to enable searching for variables to work as expected.

PR #243, opened 2018-03-22
Set auto* timers to start in 30 secs on loading, which means it will …

original-birdman:
…start its repeating countdown from when enigma2 starts each time rather then waiting until anything left over from the last enigma2 running.


original-birdman:
Local variables `time` (and `code`) in `createTimer()` were renamed to avoid a name clash with the `time()` function

PR #249, opened 2018-03-31
Add an internal function to do the bouquet search in one piece of code.

original-birdman:
This also means the test for the end of the bouquet list gets done for _both_ locations. The second one has now seen this problem too.


Huevos:
Is this code specific to OpenViX?

original-birdman:
> Is this code specific to OpenViX?

Yes. At least it's not in OpenPLi.

Huevos:
Thanks.

PR #250, opened 2018-03-31
Add another check for running out of bouquet info.

original-birdman:


original-birdman:
This also "fixes" the code in `RecordTimer.py` to use `bouquet.valid()` rather than checking the numeric type, to align with a few other instances of the check.

original-birdman:
This is Vix-specific too (not in OpenPLi).

PR #252, opened 2018-04-02
Replace the now-missing buildBouquetID() with a short piece of inline…

original-birdman:
… code.

### DIFF
--- a/lib/python/Screens/ChannelSelection.py
+++ b/lib/python/Screens/ChannelSelection.py
@@ -50,6 +50,7 @@ from Plugins.Plugin import PluginDescriptor
 from Components.PluginComponent import plugins
 from Screens.ChoiceBox import ChoiceBox
 from time import localtime, time, strftime
+import re
 try:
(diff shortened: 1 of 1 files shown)


Huevos:
Why is buildBouquetID() missing?

Huevos:
Hmm. See it here now.
https://github.com/OpenViX/enigma2/commit/01f676b4151aa5776b0d01a22a787d5f8b65dfd9#diff-cc2409666a500ee7b6bc1051460ae5e0

original-birdman:
> Why is buildBouquetID() missing?

Because a PR was committed hat removed it along with its one other usage.
This just restores the original code, albeit in-line.
It might be that the unicode code could be used here, but it at least seemed importantt to prevent the crash first.

Huevos:
Thanks Gordon. In the end I used the unicode suggestion.

https://github.com/OpenViX/enigma2/commit/e77e622abe612d581ebcdb144b5ea7ed98f67b1b

PR #254, opened 2018-04-12
Check for multiple frontends on USB tuners and use the "best" one.

original-birdman:


original-birdman:
This was found on the AstroMeta tuner.  frontend0 is a DVB-T pone, while frontend1 is DVB-T2. You want to use the latter.
It may also apply to other USB tuners. The code is generic in that it checks for frontend capability, not specific vendor.

original-birdman:
I missed changing a `frontend0` debug printing to be `frontend%d...best_fei`.  Resubmitting a fixed version.

PR #256, opened 2018-04-29
Check for multiple frontends on USB tuners and use the "best" one.

original-birdman:


original-birdman:
This was found on the AstroMeta tuner. `frontend0` is a DVB-T one (rtl2832u chipset), while `frontend1` is DVB-T2 (mn88473 chipset). You want to use the latter.
It may also apply to other USB tuners.
The code is generic in that it checks for the frontend capability, not a specific vendor.

PR #263, opened 2018-05-27
[config.py] PEP8 clean up

IanSav:
This PEP8 clean up is the first step in sequence of planned changes to fix, enhance and tidy up this code.


IanSav:
By the way, this pull request should make future merges from OpenPLi easier as they have already merged in the PEP8 clean up.

PR #268, opened 2018-06-14
Allow multiple genre values to be displayed in event information

prl001:
Fixes a bug that prevented access to genre & parental ratings lists from
Python.

Adds GenreList conversion type to EventName to support multiple
genres available in IceTV EPG (Australian non-broadcaster EPG provider).

Adding the GenreList conversion type simplifies keeping the OverlayHD
skin portable between Beyonwiz and OpenViX.

Genre conversion type in EventName now implemented on the genre list to
allow more flexibility in chosing a single genre from those available.

Added SeparatorSlash and SeparatorComma to separate items in genre list
display (defaults to SeparatorComma).

Add new methods PyObject *eServiceEvent::getGenreDataList() const and
PyObject *eServiceEvent::getParentalDataList() const so that their
Python/SWIG API signatures are differentiated from
RESULT eServiceEvent::getGenreData(ePtr<eGenreData> &dest) const and
RESULT eServiceEvent::getParentalData(ePtr<eParentalData> &dest) const

Add
PyObject *eServiceEvent::getComponentDataList() const
for consistent naming.

Re-implement
PyObject *getComponentData() const,
PyObject *getGenreData() const
and PyObject *getComponentData() const
as deprecated inline methods that call the respective new methods.

Add new GenreList type to the EventName converter to allow skin
access to the new genre list retrieval method.

The Rating types for the EventName converter have been left unchanged
because the use of a ratings list is more complicated is no way to
test the code for multiple country identifiers in Australian EPGs.

### DIFF
--- a/lib/python/Components/Converter/EventName.py
+++ b/lib/python/Components/Converter/EventName.py
@@ -95,6 +95,7 @@ class EventName(Converter, object):
 	PDCTIME = 12
 	PDCTIMESHORT = 13
 	ISRUNNINGSTATUS = 14
+	GENRELIST = 15
 
 	NEXT_DESCRIPTION = 21
 	THIRD_NAME = 22
@@ -119,6 +120,7 @@ class EventName(Converter, object):
 		"NextNameOnly": ("type", NAME_NEXT2),
 		"NameNextOnly": ("type", NAME_NEXT2),
 		"Genre": ("type", GENRE),
+		"GenreList": ("type", GENRELIST),
 		"Rating": ("type", RATING),
 		"SmallRating": ("type", SRATING),
 		"Pdc": ("type", PDC),
@@ -135,6 +137,8 @@ class EventName(Converter, object):
 		# Options...
 		"Separated": ("separator", "\n\n"),
 		"NotSeparated": ("separator", "\n"),
+		"SeparatorSlash": ("separator", "/"),
+		"SeparatorComma": ("separator", ", "),
 		"Trimmed": ("trim", True),
 		"NotTrimmed": ("trim", False)
 	}
@@ -151,7 +155,7 @@ class EventName(Converter, object):
 		self.epgcache = eEPGCache.getInstance()
 
 		self.type = self.NAME
-		self.separator = "\n"
+		self.separator = None
 		self.trim = False
 
 		parse = ","
@@ -163,6 +167,9 @@ class EventName(Converter, object):
 				print "[EventName] ERROR: Unexpected / Invalid argument token '%s'!" % arg
 			else:
 				setattr(self, name, value)
+		if self.separator is None:
+			default_sep = "SeparatorComma" if self.type == self.GENRELIST else "NotSeparated"
+			self.separator = self.KEYWORDS[default_sep][1]
 
 	def trimText(self, text):
 		if self.trim:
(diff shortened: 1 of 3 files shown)


Huevos:
Are multiple genres part of DVB SI standard?

Do you know of any European provider that uses this system?

prl001:
> Are multiple genres part of DVB SI standard?

Yes, they are. ETSI EN 300 468 V1.14.1 (2014-05) Section 6.2.9 Content descriptor allows for up to 127 content identifiers in a Content Descriptor. Similarly for multiple parental ratings, in 6.2.28 Parental rating descriptor (up to 63 country code/rating pairs).

They are also allowed in Australian standard AS4599 "Digital television—Terrestrial broadcasting
Part 1: Characteristics of digital terrestrial television transmissions", and no limitations are placed in them in the Australian FreeTV Operational Practice OP39 - DVB Content Descriptor.

The enigma2 code manages multiple content identifiers right up to the interface with the Python UI code. There's a bug in the way that the single-identifier and identifier-list versions of ```eServiceEvent::getGenreData()``` interface with Python, so that the two C++ member functions have the same signature as Python methods, and so Python can only access the single-identifier C++ member function. By renaming the C++ member functions (and retaining the old C++ member function as well), both the single-identifier and identifier-list access paths are available to Python. The single-identifier method is unchanged as ```getGenreData()``` and identifier-list method is now accessible in Python as ```getGenreDataList()```.

```Converters.EventName``` has been updated so that skins can also access either a single genre value (unchanged as ```Genre```) or a list (as ```GenreList```).

```Converters.EventName``` still respects the ```config.usage.show_genre_info``` setting for people who don't want to display genre information at all.

Multiple parental ratings can also now be accessed in the Python code, though I haven't made use of that, because I have no multi-country parental ratings descriptions to test on. European developers who do have such multi-country ratings data available might want to use the capability.

> Do you know of any European provider that uses this system?

What system? The ETSI DVB standards allow for multiple genres and multiple parental ratings. I said in the commit notes that it's to support "multiple genres available in IceTV EPG (Australian non-broadcaster EPG provider)," so that need is specifically Australian.

The reason I made the pull request is at the request of IanSav, so that OverlayHD can display genres seamlessly in both the Beyonwiz and the OpenViX code, by using ```GenreList``` converters. ```GenreList``` will display multiple genres for Beyonwiz IceTV users and single genres for Beyonwiz FTA EPG users (or none on shows that don't carry any genre information). While the standards allow Australian FTA broadcasters to send multiple genres, I've only ever seen them send either one or none.

OverlayHD will work in the same way in Europe. Other skins can leave their converters as ```Genre``` and that will continue to work in the same way as it currently does, so there are no consequences for other skin designers.

I have deliberately not included the changes to ```eEPGCache::submitEventData()``` that support loading multiple genres and multiple parental ratings from external EPG providers into the EPG cache. I'm happy to make a pull request to make those changes available if there's an interest, but I rather suspect there won't be. The ```eServiceEvent```/```Converters.EventName``` changes will work without the ```eEPGCache``` changes.

IanSav:
As Prl noted this submission is at my request and will allow greater transparency between the images.  The fact that it corrects and enhances the existing code should allow for easier adoption should and other markets use, or want to use, these DVB standard options.

PR #272, opened 2018-06-26
Add tuner info to the log at start and stop of recordings.

original-birdman:


original-birdman:
Also fixed a few typos - mostly in my own, earlier comments.

original-birdman:
> Why does this need to be so verbose?

In what way is it verbose?
It used to say "start recording"; it now says "start recording on tuner: X".
In terms of lines printed it's actually more terse, as if there is an error it just reports the error line.

original-birdman:
And if you're referring to the _code_ changes then:

- The logging is done twice, so it makes sense to drop it into a function - so 1 line for the function definition.
- There are 3 if statements, and python requires at least 4 lines for an if statement with else clauses (and we do need the info from both branches) - so 12 lines there.
- We need to fetch two values - the subjects of the inner two if/else clauses - so 2 lines for these.
- 1 line for the logging
- 1 line for the import of string. 

This is all needed to provide the relevant information.

The rest of it is fixes for typos (as already noted), since the file is being modified anyway.

ccs-ccs:
Verbose or not, I'm finding it very useful trying to establish if a tuner is faulty.

If 2 recordings start at the same time it is impossible to predict which tuners will get used when they are all of the same type.

Huevos:
e.g. tuner_info = string.ascii_uppercase[tn], which also needs "string" imported.

The established way to do that in enigma is...
chr(ord('A') + slot)

But what I really meant is using a lot of code to do so little.

original-birdman:
I'll resubmit one using chr + ord, but the size of the rest of it is a limitation of the API between python and the C++ object info, coup;led with the fact that there is no guarantee that the attrs are present, so you have to cater for their absence too.

original-birdman:
So what about this?
   https://github.com/OpenViX/enigma2/pull/275

PR #275, opened 2018-06-27
Add tuner info to the log at start and stop of recordings. (v2)

original-birdman:


original-birdman:
Since you've just cherry-picked the fixes to typos in the comments (ignoring the one in actual code) I take it that _you've_ decided that recording the tuner used for a recording is just an irrelevance?

Huevos:
It is useful debug data but I would just prefer that the code followed the format used elsewhere. Right now I am just running some recordings to test.

original-birdman:
The only similar code (to get tuner data) is in `TimerSanityCheck.py` and this does follow that format, as I based it on that instance.

Huevos:
That code needs looking at too then if is reinventing the wheel.

original-birdman:
It's doing something slightly different (as it's just simulating the tuner assignment).
Not sure what wheel it's re-inventing - this seems to be the only way to get the tuner assigned to a given recording.

Huevos:
I've add this: https://github.com/OpenViX/enigma2/commit/018d63a4ff112d6bc4727b6c5e720b138b0b01ff

Please test.

original-birdman:
That works, thanks.
Doesn't distinguish between the various possibilities for not getting a tuner, but that probably doesn't matter at the moment.

Huevos:
> Doesn't distinguish between the various possibilities for not getting a tuner

That's kind of debug for debug. Anyway so far I haven't seen any "?" mark.

Had it running for a couple of hours then the electric supply went off.. :(

original-birdman:
> Anyway so far I haven't seen any "?" mark.

You'll get one if you record a stream.

Huevos:
As long as there is no crash. ;)

original-birdman:
> As long as there is no crash. ;)

The fix to stop one when you stop an instant recording is in Dev.
PS: What python _really_ needs is a class in which all method calls return an instance of itself, all tests return False and all use as a number returns 0. (So an enhanced version of None).
the you could just say:
```
tn = self.record_service.frontendInfo().getFrontendData().get("tuner_number")
tuner_info = tn and chr(ord('A') + tn) or "?"            
```

Huevos:
Yep or use a try/except. But a lot of people frown on that.

Current style is used in many places. e.g...
https://github.com/OpenViX/enigma2/blob/Dev/lib/python/Components/Sources/FrontendInfo.py#L55-L57

Huevos:
```
try:
	tuner_info = chr(ord('A') + self.record_service.frontendInfo().getFrontendData().get("tuner_number"))
except:
	tuner_info = "?"
```

But still 4 lines of code so not really gaining anything. And not as easy to read.

original-birdman:
Agreed.
I really was thinking of an entity like None, but where you could use it with any method call (just returns another None), with any integer add/subtract  (always treated 0) and in any Boolean statement (always treated as False). It would simplify a lot of coding.

original-birdman:
> I really was thinking of an entity like None, but where you could use it with any method call (just returns another None), ....

Tuns out that I'm not the only one with  that thought:
https://www.safaribooksonline.com/library/view/python-cookbook/0596001673/ch05s24.html

PR #277, opened 2018-07-25
Check for tn being "int" rather than "not None".

original-birdman:


original-birdman:
To prevent a crash for IPTV recording of:
```
<   396.082> Traceback (most recent call last):
<   396.082>   File "/usr/lib/enigma2/python/timer.py", line 239, in calcNextActivation
<   396.082>   File "/usr/lib/enigma2/python/timer.py", line 314, in processActivation
<   396.083>   File "/usr/lib/enigma2/python/RecordTimer.py", line 938, in doActivate
<   396.083>   File "/usr/lib/enigma2/python/RecordTimer.py", line 613, in activate
<   396.083>   File "/usr/lib/enigma2/python/RecordTimer.py", line 440, in log_tuner
<   396.084> TypeError: unsupported operand type(s) for +: 'int' and 'dict'

```

Huevos:
Please explain at what time is this something other than None or an int?

original-birdman:
See:
https://www.world-of-satellite.com/showthread.php?59774-IPTV-timers-crash&p=472618#post472618

That excerpt is from the log in post 10

Huevos:
But this is just code is just hiding the problem. What needs to be looked at is why the underlying code is returning an incorrect type.

original-birdman:
> But this is just code is just hiding the problem.

Possibly. But in the meantime this fixes the crash (hence enabling the end user to actually do a recording, which is the most important thing).
It's also more correct code, since you only want to do the addition if you have an `int`.

Huevos:
Sorry, I don't agree. If fedata["tuner_number"] contains something other than an int that needs investigating.

RobvanderDoes:
Not sure if this is related, but I get crashes every now and again manually adding a DVB-S record timer (from single-EPG). There is no recipe: most of the times there is no problem and then out of the blue the odd crash.
[Enigma2_crash_2018-07-26_17-48-39.zip](https://github.com/OpenViX/enigma2/files/2234343/Enigma2_crash_2018-07-26_17-48-39.zip)

original-birdman:
> Not sure if this is related,...

It's not*.
That crash has actually occurred when the AutoTimer background checker ran, tried to add a timer for Luther and failed to find the file to rename after editing.
I suggest you open a separate bug report about it.

* So it is related in a way - as this is another event which "cannot happen". The file was opened and written to without error in the preceding lines, so how can it not be there now?

PR #282, opened 2018-07-30
[Standby.py] Create names for exit values

IanSav:
Create standard names for the numeric exit codes.  This provides clarity of meaning for the exit codes and allows external code to more logically request various restart modes without needing to use magic numbers.

### DIFF
--- a/lib/python/Screens/Standby.py
+++ b/lib/python/Screens/Standby.py
@@ -17,6 +17,14 @@ import os
 
 inStandby = None
 
+QUIT_SHUTDOWN = 1
+QUIT_REBOOT = 2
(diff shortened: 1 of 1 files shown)


andyblac:
numbers are used for init values, ie linux init

IanSav:
I think code is more readable if you write something like:
`self.session.open(TryQuitMainloop, retvalue=QUIT_RESTART)`
rather than:
`self.session.open(TryQuitMainloop, retvalue=3)`

andyblac:
all I was saying is unless you are willing to go through the whole e2 code base and change the necessary code, it is not much use.

PR #283, opened 2018-08-04
fileHas - implement in SystemInfo audio/video checks and use to reduc…

TwolDE2:
…e code in AVSwitch
This uses fileHas introduced in last Dev commit and translates OpenPli changes to move File checks into SystemInfo and cleanup AVSwitch code, also simplifies a couple of SystemInfo elements.
Tested on my Arm receivers.

### DIFF
--- a/lib/python/Components/AVSwitch.py
+++ b/lib/python/Components/AVSwitch.py
@@ -557,16 +557,7 @@ def InitAVSwitch():
 	iAVSwitch.setInput("ENCODER") # init on startup
 	SystemInfo["ScartSwitch"] = eAVSwitch.getInstance().haveScartSwitch()
 
-	if os.path.exists("/proc/stb/hdmi/bypass_edid_checking"):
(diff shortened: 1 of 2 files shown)


Huevos:
Can you point out which PLi commits this relates to so we can skip these from anything we merge. Looks like the plugin commits related to this still need merging.

TwolDE:
OpenPli: 
Directories.py 9728751 is in Dev (adds fileHas)
SystemInfo.py 125b383 & 25c81cd ..... coupled to AVSwitch 32688d3, 

but OpenPli use different SystemInfo element names to ViX, so this pull uses the ViX AV elements as implemented across AVSwitch.py, AudioSelection.py & VideoMode.py (and possibly others), so it was safer to keep the same ViX naming for the AV elements, rather than changing everything to the OpenPli names.
Also the PCMmultichannel code is more complex than in OpenPli so had to be handled differently.
................. Andy implemented most of this AV AVSwitch stuff before OpenPli.

PR #296, opened 2018-10-09
Sf8008

TwolDE2:
The SystemInfo change needs the canMultiBoot definition before the change in hasmmc it would appear . I guess logical.


Huevos:
Twol, looks like this conflict with #295

PR #300, opened 2018-10-24
Movielist sort by duration

SimonCapewell:
- Added sorting movielist by duration followed by alpha
- Also resurrected the duration column which looks to have been in a previous version. Makes use of the existing setting that currently appears to have no effect. Any suggestions how to make this more language friendly? Pick it up from a setting? It's hard coded to **X Mins** as things stand.
- Fixed issue 299


Huevos:
Sorry but all PRs go to Dev branch, not master.

PR #310, opened 2018-11-04
Graph epg selection bugfix

SimonCapewell:
Fixes a weird double selection rect glitch when opening the grid EPG if you have any setting other than servicename

Also adds an option to show servicenumber+picon. I'm not keen on the looks of picon then servicenumber: it looks a bit clunky. The channel list does it better, i.e. number, picon, name. Any thoughts on swapping this around in the EPG?

### DIFF
--- a/lib/python/Components/EpgList.py
+++ b/lib/python/Components/EpgList.py
@@ -307,8 +307,6 @@ class EPGList(GUIComponent):
 		self.showServiceNumber = "servicenumber" in value
 		self.showServiceTitle = "servicename" in value
 		self.showPicon = "picon" in value
-		self.recalcEntrySize()
-		self.selEntry(0) #Select entry again so that the clipping region gets updated if needed
 
 	def setOverjump_Empty(self, overjump_empty):
 		if overjump_empty:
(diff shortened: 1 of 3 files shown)


Huevos:
Please describe the bug that this is to fix.

SimonCapewell:
Sure, you should be able to see this yourself by following these steps, it's not skin dependent
1. Set your graph or text EPG to picon only or anything other than just "service name"
2. Open the graph or text EPG. The code without bugfix does 2 calculations of selection rectangle
    1st is during fillGraphEPG before *setShowServiceMode* has been called, which means it uses the default setting of *servicename*
    2nd time is during *setShowServiceMode*, which recalculates everything again. 

3. When you move the selection down a row, the correct selection rect gets removed, but the incorrect rect remains highlighted:
nd

4. Wait 10 seconds or so and the autorefresh clears the erroneous selection artifact

SimonCapewell:
> May break infobar graph epg.
> As it now ignores the title preferance config.epgselection.infobar_servicetitle_mode.value

Yes, quite right. Overeager code simplification. I've rectified this.

>Also it will not update the list on a change

Yes, however, unless I'm missing something, *EPGSelection.onCreate* is the only caller and it's immediately doing an update in *fillGraphEPG*. If it's a point of principal to have the update just in case another caller is going to use the component, I'd be inclined to introduce a parameter to say not to update while the component is being initialised before first render.

RobvanderDoes:
I confirm bug and fix.

RobvanderDoes:
I confirm this (cosmetic) bug, and fix works without side effects noticed.

PR #312, opened 2018-11-04
Honour autotime startup delay

SimonCapewell:
Autotimer has a setting for startup delay, which was not being used by the timezones module, thus the setting was effectively ignored. Changed the const to pick up the value from config.plugins.autotimer.delay


SimonCapewell:
Bah. This wasn't supposed to have the EPG branch changes in it either. Closing

PR #313, opened 2018-11-04
Honour autotimer startup delay setting

SimonCapewell:
Autotimer has a setting for startup delay, which was not being used by the timezones module, thus the setting was effectively ignored. Changed the const to pick up the value from config.plugins.autotimer.delay


SimonCapewell:
^ Wrong PR :-) I'll answer this in PR 310

PR #314, opened 2018-11-04

original-birdman:
in "global sort method" mode.  Without this the blue "Sort by" options
do nothing - not even say they will do nothing.

### DIFF
--- a/lib/python/Components/MovieList.py
+++ b/lib/python/Components/MovieList.py
@@ -10,7 +10,7 @@ from Components.config import config
 from Tools.LoadPixmap import LoadPixmap
 from Tools.Directories import SCOPE_ACTIVE_SKIN, resolveFilename
 from Screens.LocationBox import defaultInhibitDirs
-#GML:1
+#
 from Tools.Trashcan import getTrashFolder
 import NavigationInstance
 import skin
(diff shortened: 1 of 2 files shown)


SimonCapewell:
I've been caught out by this when I had "individual settings for each directory" set to no. I thought sorting using the blue button was broken! The behaviour I was expecting at the time was for the sort to be remembered, i.e. blue being a shortcut to setting the global sort method.

original-birdman:
I reckon it's more useful to do a one-off sort with a different method.  I normally look for things by name, but can also envisage wanting to find a film which is less then 2 hours long.

This change does put up text to remind you which is happening.

The global sort method, and whether it is on, are in the Settings.

Huevos:
Birdman please can you post a method to replicate the bug you are fixing.

original-birdman:
You can replicate it from the info in the commit comment.

Disable the "sort method per-directory" option (so that you have the same sort method for all locations) then open up a movie directory and press the blue "Sort by". Regardless of what you then sleect, nothing happens.

See:
https://www.world-of-satellite.com/showthread.php?59298-Sort-order-in-recording-list-not-working


However I need to re-submit this, as I forgot to put the _(..) i18n bits around the text messages.

Huevos:
OK so we close this one then until you are ready.

original-birdman:
Re-done as https://github.com/OpenViX/enigma2/pull/315

PR #315, opened 2018-11-05

original-birdman:
in "global sort method" mode. Without this the blue "Sort by" options
do nothing (not even say they will do nothing).

### DIFF
--- a/lib/python/Components/MovieList.py
+++ b/lib/python/Components/MovieList.py
@@ -10,7 +10,7 @@ from Components.config import config
 from Tools.LoadPixmap import LoadPixmap
 from Tools.Directories import SCOPE_ACTIVE_SKIN, resolveFilename
 from Screens.LocationBox import defaultInhibitDirs
-#GML:1
+#
 from Tools.Trashcan import getTrashFolder
 import NavigationInstance
 import skin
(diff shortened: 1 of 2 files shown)


original-birdman:
This is a re-submit of https://github.com/OpenViX/enigma2/pull/314 with i18n added to the two additional text messages.

RobvanderDoes:
Confirmed bug.
But fix is not complete: although the global sorting now works, the selected sort remains showing the old value.
e.g. sorting was 'alfa then oldest', change it to 'by date'. The result is indeed the list being sorted by date, but . alfa the oldest' still shows as selection.

original-birdman:
>> The result is indeed the list being sorted by date, but 'alfa then oldest' still shows as selection.

That's because that will still be the selection. It's a temporary sort-change just for that one view of that one directory.
Or are you saying that if this temporary sort is set then this should be the method highlighted when you  press "Sort by" again?  I suppose that might make sense so that you can find out what is the current sort method. Hmm.....thinking....

RobvanderDoes:
Yes, that's what I mean, and that's how it works when the "sort method per-directory" is enabled.

original-birdman:
OK. Thinking done. Now testing...

original-birdman:
https://github.com/OpenViX/enigma2/pull/316

PR #316, opened 2018-11-08

original-birdman:
in "global sort method" mode. Without this the blue "Sort by" options
do nothing (not even say they will do nothing).


original-birdman:
This highlights the _current_ sort method in the "Sort by" list.

RobvanderDoes:
Thanks, works fine.

PR #318, opened 2018-11-10
Honour autotime startup delay

SimonCapewell:
Fix for last PR that didn't actually fix the bug. Python caught me out with its dynamic variables and requirement to having to prefix with *self.*


Huevos:
Sorry, this commit has a merge conflict.

SimonCapewell:
That merge should sort it

PR #325, opened 2018-11-16
Movielist picons

SimonCapewell:
I like picons. Here's some in the movielist, configurable of course. When enabled, they appear to the left of the recording name and the progress icon moves to the right of the recording name, Useful in the UK where some channels have loads of adverts, others have none,


RobvanderDoes:
I'm not sure if this is a good idea; opening the movie list takes ages when this PR is taken aboard, even if showing the picons is set to no (tested on 5.2.014.005 developer).

RobvanderDoes:
Retested and not able to reproduce the issue, hence merged.

SimonCapewell:
If it's any help, I've been running this patch or something quite close to it on my ET8500 for 6 months or so and not noticed any poor performance

PR #326, opened 2018-11-17
auto suggest new folder name from selection

SimonCapewell:
2 enhancements to the movie list screen

- In the movielist screen, adds ability to assign a shortcut key to create directory
- When creating a new directory, the virtual keyboard opens with the name of the selected programme as a suggestion. This make it much easier to group a bunch of programmes.

### DIFF
--- a/lib/python/Screens/MovieSelection.py
+++ b/lib/python/Screens/MovieSelection.py
@@ -753,6 +753,7 @@ class MovieSelection(Screen, HelpableScreen, SelectionEventInfo, InfoBarBase, Pr
 				'copy': _("Copy"),
 				'reset': _("Reset"),
 				'tags': _("Tags"),
+				"createdir": _("Create directory"),
 				'addbookmark': _("Add bookmark"),
 				'bookmarks': _("Location"),
 				'rename': _("Rename"),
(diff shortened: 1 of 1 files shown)


AbuBaniaz:
Would it be possible to focus energy on moving the tags into the sort options. It is currently activated by the radio button.

Being able to sort by tags easily/intuitively will achieve similar functionality to using folders.

RobvanderDoes:
> Being able to sort by tags easily/intuitively will achieve similar functionality to using folders.

I agree 100%.
I have achieved that (for years already) by assigning that function to one of the colored buttons (the movie list context menu allows for that).

SimonCapewell:
Interesting suggestion. Would you care to expand on how you'd see the sort menu working in that case? 

Tags still fall short in a few scenarios:
1. User folders. We have one per person for the stuff we know nobody else will be interested in. I can't see a way currently to make custom tags to achieve the same end, especially with one off timers.
2. Genre categorising. We have comedy, drama, documentaries, family, films. Same as 1 really. You could perhaps have the tags capture the broadcast EPG genre, but that leaves you at the mercy of the broadcasters, i.e. the current infuriating fashion for prefixing things randomly with *New:!!! !!LOOK!! !!AT!! !!ME!! The One Show*
3. Going for a flat directory structure might impact the opening time of the movielist as more files have to be scanned. 
4. Given 1, filters per folder may be a prudent thing to add. The application of a filter isn't blazingly obvious, it's in the title bar, but in a household containing non techy people panic will ensue
over "missing recordings".
5. Some of my suggested tags have underscores instead of spaces. For example, I've got *The_IT_Crowd* and *The IT Crowd*. I've also got *The* as a suggestion.

ccs-ccs:
@SimonCapewell, regarding point 5 in your post, could this be the reason? …..

`https://www.world-of-satellite.com/showthread.php?59929-Tags-not-working-quite-as-expected&p=473897&viewfull=1#post473897`

IanSav:
Given that most parsers break on white space it is common for all systems, not just Unixy ones ;), to have problems with filenames with spaces in them.  There are many areas in Enigma2 that fail to properly protect filenames that include spaces.  Using underscores to replace spaces in file names ensures that those areas of the code that do not protect against spaces won't break.

While it would be great to to fix all filename handling in Enigma2 it is probably wise to stick with filenames that don't break.  :)

original-birdman:
There's a bug in this at line 1827 (as was):
      `dirname, self.extension = os.path.splitext(name)`_
but there is no variable called `name`.
I _suspect_ this should be:
      `dirname, self.extension = os.path.splitext(full_name)`

SimonCapewell:
Thanks, you're almost right, it should be `dirname`

original-birdman:
Given that the code only gets executed if `dirname == full_name` I reckon either would suffice.

PR #328, opened 2018-11-20
NetworkSetup - reinstate MAC changeability

TwolDE2:
The ability to change the MAC address in NetworkSetup was merged from OpenATV by Andy on Jan 3rd 2014 with some modifications for OpenViX - one of which was to insert a kernel check, which I guess was valid at the time - but has never been changed.
This small change re-enables the code - checked out on my SF8008 and appears fine.

### DIFF
--- a/lib/python/Screens/NetworkSetup.py
+++ b/lib/python/Screens/NetworkSetup.py
@@ -1156,9 +1156,7 @@ class AdapterSetupConfiguration(Screen, HelpableScreen):
 
 		if os_path.exists(resolveFilename(SCOPE_PLUGINS, "SystemPlugins/NetworkWizard/networkwizard.xml")):
 			menu.append((_("Network wizard"), "openwizard"))
-		kernel_ver = about.getKernelVersionString()
-		if kernel_ver >= "3.5.0":
-			menu.append((_("Network MAC settings"), "mac"))
+		menu.append((_("Network MAC settings"), "mac"))
 
 		return menu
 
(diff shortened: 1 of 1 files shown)


RobvanderDoes:
The reason for this statement at the time was that the ability to change the address had been removed from newer kernels (at least on a number of brands). So I'm not sure if we can delete this check now for all brands.

TwolDE:
People are doing this manually regardless of brands, so maybe a warning message -? 
There are NO checks on the other images........ and many support far more receivers

RobvanderDoes:
At the time certain kernels (brands?) simply didn't allow changing the mac address. Even though the GUI seemed to do it, in reality the address wan't changed at all.
That other images don't have such a check doesn't mean that it works properly.
IMHO it should be tested if changing the address via the GUI now actually works for all kernels/brands before deleting the check.

TwolDE:
OK, sounds reasonable, so perhaps we can add to Dev and test?
Or .....? I guess its not a big deal as the users can do this manually, so perhaps just reject the pull..... or even delete the code completely?

RobvanderDoes:
I suppose it still works for older kernels.
Adding to DEV is not required for testing; that can be done by hand. But one needs enough testers & especially enough hardware.

ViX-Sicilian:
Twol, how come you closed this? I was going to accept into dev in the dev build 5.2.015.xxx

RobvanderDoes:
I only wanted to remind you guys about the history of this functionality. I had no intention for the PR to be closed.

PR #330, opened 2018-11-22
Graphepg movement algorithm

SimonCapewell:
I've noticed moving sideways within the graph EPG screen is much slower than moving up and down. It turns out the entire grid is invalidated and redrawn, compared to just the current and previous rows when moving up and down services.

This change invalidates just the current row and makes movement within the current page much more snappy.


SimonCapewell:
Build shouldn't have failed. Only 1 python file changed :-/

PR #336, opened 2018-11-29
Use the correct variable in test (sv, not the non-existant val).

original-birdman:


original-birdman:
Clearly this code is little used, but when it is enigma2 crashes...

PR #347, opened 2018-12-08
Fix Graphical-EPG to allow prev-page after +24 hours.

original-birdman:
General tidying-up of some code as well.


original-birdman:
Contains fix for single-EPG crash.

PR #349, opened 2018-12-10
[ButtonScreen] Efficiency issue in extracting help info

prl001:
Improve average speed of InfoBarButtonSetup.getKeyFunctions() when used
when extracting an ActionMap's help information by only searching the
getButtonSetupFunctions() list if the strings in
config.misc.ButtonSetup.<keyname> are non-empty.

In all cases where config.misc.ButtonSetup.<keyname> is used,
use getattr() to get the relevant config entry instead of using eval over a constructed string, saving the need to compile the expression.

Where the list returned getButtonSetupFunctions() is being searched for
a match, avoid constructing a list of all matches and terminate early
by using next(...generator..., None) instead of generating a list from a
full search.

### DIFF
--- a/lib/python/Screens/ButtonSetup.py
+++ b/lib/python/Screens/ButtonSetup.py
@@ -252,10 +252,10 @@ class ButtonSetup(Screen):
 		key = self["list"].l.getCurrentSelection()[0][1]
 		if key:
 			selected = []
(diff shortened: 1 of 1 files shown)


RobvanderDoes:
PR is against 'Master'. Please re-submit against dev-branch.

RobvanderDoes:
I picked the PR in the Dev-branch.

PR #365, opened 2019-01-15
multiboot - directly mount boot partition - ignore fstab mount

TwolDE2:
Users are restoring backup settings from other receivers which means that the fstab is replaced and contains invalid information for multiboot receivers where fstab contains the boot partition - and leads to crashes or other issues.
This change plus a separate ImageManager change removes the dependency by mounting the boot partition directly before changing STARTUP files (in boot)

### DIFF
--- a/lib/python/Screens/MBRestart.py
+++ b/lib/python/Screens/MBRestart.py
@@ -1,12 +1,15 @@
-from Screens.Screen import Screen
-from Screens.Standby import TryQuitMainloop
-from Screens.MessageBox import MessageBox
+from os import mkdir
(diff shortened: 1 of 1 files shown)


Huevos:
Just got to wait for Stef to confirm his release build is complete before pulling this.

ViX-Sicilian:
Will be another 2-3 days I think. I'll update you all once complete.

TwolDE:
No issue - I expected the delay, Just thought I would get it out of the way once I had decided that the best solution was to ignore the OE-A  “fudge“ with the fstab and just mount the volume.

PR #367, opened 2019-01-20
Update Timeshift.py

bbbuk:
Add support for pause/play IPTV stream instead of displaying error message (or not, depending on settings) stating timeshift not supported.

This isn't full timeshift (ie seek) support with IPTV streams, rather a pause/playback.  Maybe over time, this could be incorporated if possible!

Many thanks to @danny121pt for posting original mod and the two timeshift commits on ATVs implementation of this.

Tested working on solo2 with normal dvb streams and IPTV 4097: streams (iptv 1: stream had few issues but this could have been my box as 4097 is more stable IPTV stream on my box anyhow).


Huevos:
@bbbuk please submit this PR to Dev branch, not master.

PR #368, opened 2019-01-20
Update timeshift.py - introduce pause/play on IPTV

bbbuk:
This adds support for pause/play of an IPTV stream instead of displaying error message (or not, depending on settings) stating timeshift not supported.

This isn't full timeshift (ie seek) support with IPTV streams, rather a pause/playback.  Maybe over time, this could be incorporated if possible!

Many thanks to @danny121pt for posting original mod and the two timeshift commits on ATVs implementation of this.

Tested working on solo2 with normal dvb streams and IPTV 4097: streams (iptv 1: stream had few issues but this could have been my box as 4097 is more stable IPTV stream on my box anyhow).

### DIFF
--- a/lib/python/Components/Timeshift.py
+++ b/lib/python/Components/Timeshift.py
@@ -47,7 +47,7 @@ from Tools.BoundFunction import boundFunction
 from Tools.Directories import pathExists, fileExists, getRecordingFilename, copyfile, resolveFilename, SCOPE_TIMESHIFT
 from Tools.TimeShift import CopyTimeshiftJob, MergeTimeshiftJob, CreateAPSCFilesJob
 
-from enigma import eBackgroundFileEraser, eTimer, eServiceCenter, iServiceInformation, iPlayableService, eEPGCache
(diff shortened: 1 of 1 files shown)


RobvanderDoes:
Please give this a thorough test; also using FallBack tuner with only 10mbs LAN-connection.

bbbuk:
tbh, I didn't know there was an issue when using fallback tuner.  This PR has been in atv for a year so I would like to think if it caused major issues it would have been reverted by now.

Worst case is, maybe a yes/no is added in menu for timeshift?  Allow user to choose whether they want to allow play/pause on IPTV.  If user chooses no (can be default), then nothing happens just like it was before.

This way, if needed, it puts choice on users' as they may not use fallback but would miss out on pause/play an IPTV stream because of a feature they may not use.

Huevos:
We are not going to add a user configurable option as a fix for a BSoD.

Try adding...
self.pts_trycount = 0
in the _init_

PR #370, opened 2019-01-22
Movielist column widths

SimonCapewell:
Movielist enhancements

- Columns size proportionally to the font size
- Directory icon no longer shows if progress and picons are disabled, making more space for titles
- Progressbar/progress icon and picons vertically aligned with text
- Option added to context menu to show/hide the duration column


AbuBaniaz:
>>config.usage.load_length_of_movies_in_moviellist

Is that a typo?

RobvanderDoes:
One more question: can you also move the "Show status icons in movie list" from the recordings setup to the MovieList context menu please?

SimonCapewell:
1. Could do, it might look a bit messy though. That'd be picon, progress, name?
2. IIRC, the code for the option was already in there when I started, just the display of the column was disabled in movielist.py. If it's non standard, we should change it.
3. The option's there, about 1/3 of the way down the context menu. Perhaps this demonstrates that the ordering of the menu isn't the best!

RobvanderDoes:
> One more question: can you also move the "Show status icons in movie list" from the recordings setup to the MovieList context menu please?

Done in https://github.com/openvix/enigma2/commit/5e52b789b718eceba75f889d36c917ad1abce9b8 as it was also a double entry.

> Could do, it might look a bit messy though. That'd be picon, progress, name?

Yes, that's the idea.
TBH: IMHO the current situation is a bit messy, see:
ml-07

SimonCapewell:
Yes, looks like `load_length_of_movies_in_moviellist` was a typo from a long time ago. It only appears in the config file if you've set it to off. We could change it - I'm unclear what the convention is for config naming and location, would `config.movielist.load_length_of_movies` be more appropriate?

One idea that springs to column order wise is loading the order from the skin, e.g. the skin defines `columns='picon,progress,name,duration,date'` and you get that order. 

BTW while investigating this I realised that I've rendered some skin properties obsolete, `partIconeShift` being an example. This could potentially be undesirable for any skins that want to deliberately put icons out of line of the text. Any thoughts?

RobvanderDoes:
> One idea that springs to column order wise is loading the order from the skin, e.g. the skin defines `columns='picon,progress,name,duration,date'` and you get that order.

I tried that, but that doesn't work:
`<   130.327> [Skin] Attribute not implemented: columns value: picon,progress,name,duration,date`

SimonCapewell:
That's because it's just an idea at the moment :-) I'm looking for some opinions on whether it's a desirable feature before coding it up.

RobvanderDoes:
Ah, sorry, I thought you suggested this to be implemented 😊

IanSav:
@SimonCapewell, you may care to have a look at the Beyonwiz enhancements to the MoviePlayer.  It adds things like selection lists (that include playlists) as well as more sort orders and column management.  There is probably a lot more in there but as a regular user of that version I now take the changes for granted.

[https://bitbucket.org/beyonwiz/easy-ui-4/src/dd3f1fa0cad9fe7b4fbd68045d4647aa79ce4640/lib/python/Screens/MovieSelection.py?at=master](url)
[https://bitbucket.org/beyonwiz/easy-ui-4/src/dd3f1fa0cad9fe7b4fbd68045d4647aa79ce4640/lib/python/Components/MovieList.py?at=master](url)

AbuBaniaz:
>That's because it's just an idea at the moment :-) I'm looking for some opinions on whether it's a desirable feature before coding it up.

Adding features that skinners can use, without breaking existing/other skins is a desirable feature.

@IanSav 
Those links are being auto-resolved to non-working ones. You may wish to edit them with backticks

IanSav:
Links fixed.  Thanks.

RobvanderDoes:
> Adding features that skinners can use, without breaking existing/other skins is a desirable feature.

Absolutely!

> BTW while investigating this I realised that I've rendered some skin properties obsolete, partIconeShift being an example. This could potentially be undesirable for any skins that want to deliberately put icons out of line of the text. Any thoughts?

No response from anyone sofar. Although I'm in favour of skinning flexibility, I have no idea if anyone would like to do that.

PR #372, opened 2019-01-23
[Timeshift.py] Fix false recording status/config

original-birdman:
…to be set it will already be set.


RobvanderDoes:
Not sure if related to this commit, but atm:
1- No more crash when activating TS on stream from fallback tuner (at least I wasn't able to provoke it anymore).
2- Although enabled there is no warning when activating TS on a stream.

ts-warningenabled

original-birdman:
Unless you have "Stop timeshift while recording" set (which you don't) this change will have no effect, as `ptsTimerEntryStateChange()` will return without doing anything (and always would have done so).

PR #376, opened 2019-01-28
[MovieSelection.py] Use new sort settings

IanSav:
In the sortBy() method the newly selected sort mode is not set or saved if the config.movielist.settings_per_directory is False.  This change updates config.movielist.moviesort with the newly selected sort mode and then saves that setting for future use.

### DIFF
--- a/lib/python/Screens/MovieSelection.py
+++ b/lib/python/Screens/MovieSelection.py
@@ -1444,6 +1444,8 @@ class MovieSelection(Screen, HelpableScreen, SelectionEventInfo, InfoBarBase, Pr
 			if config.movielist.settings_per_directory.value:
 				self.saveLocalSettings()
 			else:
+				config.movielist.moviesort.value = newType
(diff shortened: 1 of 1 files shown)


RobvanderDoes:
I fail to see the issue. I always have config.movielist.settings_per_directory set to true, and the sorting is always saved neatly. The result being that my sorting is indeed different per directory, and a changed sorting for a specific directory is saved.

IanSav:
The problem is that if config.movielist.settings_per_directory is set to False then the sort settings aren't used and can't be changed or saved.

Huevos:
Hi Ian, thanks for this PR. Going to test now.

RobvanderDoes:
Ah, OK, so the other way around than the commit text says :(

IanSav:
IMPORTANT NOTE: Original commit and the comment above has "config.movielist.settings_per_directory is True" but should be "config.movielist.settings_per_directory is False".

PR #377, opened 2019-01-30
hide sf8008 SDcards when used for multiboot + use correct SF8008 boot…

TwolDE2:
… partition for multiboot

Hardisk.py - simplify BLACKLIST for eMMC receivers
About.py -don't list SDcard if used for multiboot 
MBRestart - SF8008 use correct boot partition in restart

### DIFF
--- a/lib/python/Components/Harddisk.py
+++ b/lib/python/Components/Harddisk.py
@@ -4,7 +4,7 @@ from Tools.CList import CList
 from SystemInfo import SystemInfo
 from Components.Console import Console
 from Tools.HardwareInfo import HardwareInfo
-from boxbranding import getMachineBuild
+from boxbranding import getMachineBuild, getMachineMtdRoot
 import Task
 import re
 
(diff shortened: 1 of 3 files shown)


andyblac:
has this been tested with all box variations that has SDCard slots? ie just SDCard and no eMMC, both SDCard slot and eMMC, and just eMMC

TwolDE2:
The SF8008 when it formats the SDcard creates 4 partitions.
In About & MountManager it specifically tests that it is a multiboot receiver, that it is using both SDcard and mmcblk partitions  and that path /dev/sda4 actually exists before bypassing the addition of any devices to the list. 
The changes have been tested on the SF8008 with no SDcard, blank SDcard(not formatted), normal format SDcard (both FAT32 & ext4) and a multiboot (4 partitions) SDcard.

ViX-Sicilian:
Guys, I'm testing this now.

PR #379, opened 2019-02-08
Skip "go to standby?" code when  a Zap timer ends.

original-birdman:


original-birdman:
Also report the current tuner in the log when a zap timer ends.

PR #378, opened 2019-02-06
Ensure updated sort method is saved in per-directory mode.

original-birdman:
In global sort mode let the user choose whether the change is permanent or temporary.

PR #380, opened 2019-02-14
    Ensure updated sort method is saved in per-directory mode.

original-birdman:
In global sort mode let the user choose whether the change is permanent or temporary.

### DIFF
--- a/lib/python/Components/MovieList.py
+++ b/lib/python/Components/MovieList.py
@@ -148,7 +148,7 @@ class MovieList(GUIComponent):
 	HIDE_DESCRIPTION = 1
 	SHOW_DESCRIPTION = 2
 
-# So MovieSelection.selectSortby() can find out whether we are 
+# So MovieSelection.selectSortby() can find out whether we are
 # in a Trash folder and, if so, what the last sort was
 # The numbering starts after SORT_* values above.
 # in MovieSelection.py (that has no SORT_GROUPWISE)
@@ -351,7 +351,7 @@ class MovieList(GUIComponent):
 		durationWidth = self.durationWidth if config.usage.load_length_of_movies_in_moviellist.value else 0
 
 		width = self.l.getItemSize().width()
-		
+
 		dateWidth = self.dateWidth
 		if not config.movielist.use_fuzzy_dates.value:
 			dateWidth += 30
@@ -684,8 +684,9 @@ class MovieList(GUIComponent):
 		self.list.sort(key=self.buildGroupwiseSortkey)
 
 # Have we had a temporary sort method override set in MovieSelectiom.py?
-# If so use it, remove it (it's a one-off) and remember the method so
-# that the "Sort by" menu can highlight it.
+# If so use it, remove it (it's a one-off) and set the current method so
+# that the "Sort by" menu can highlight it and "Sort" knows which to
+# move on from (both in Screens/MovieSelection.py).
 #
 		try:
 			self.current_sort = self.temp_sort
--- a/lib/python/Screens/MovieSelection.py
+++ b/lib/python/Screens/MovieSelection.py
@@ -53,6 +53,7 @@ config.movielist.videodirs = ConfigLocations(default=[resolveFilename(SCOPE_HDD)
 config.movielist.last_selected_tags = ConfigSet([], default=[])
 config.movielist.play_audio_internal = ConfigYesNo(default=True)
 config.movielist.settings_per_directory = ConfigYesNo(default=True)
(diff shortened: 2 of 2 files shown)


original-birdman:
There are still some issues to look at in this area, but this fixes the one reported bug.

The further refinements are being discussed at:
https://www.world-of-satellite.com/showthread.php?60840-Issues-with-new-MovieSelection-changes&p=481559&viewfull=1#post481559
_et seq_

RobvanderDoes:
I hope others will give their opinion on the doubts I reported on saving (see https://github.com/OpenViX/enigma2/pull/378#issuecomment-463615921 under A2 and C1) before merging this PR.

IanSav:
Given that this is not in a Dev build I will try to allocate some time over the weekend to integrate the changes into my network so that I can give them a go.

RobvanderDoes:
It's so easy to download the py........

@original-birdman : I think something went wrong, as this is the same file as in the now closed PR https://github.com/OpenViX/enigma2/pull/378

IanSav:
> It's so easy to download the py........

There is a LOT more to it than a simple download in my environment.

RobvanderDoes:
> There is a LOT more to it than a simple download in my environment.

Than things seem to be over complicated in your environment. The test is nothing more than the last dev-build + one downloaded py.

IanSav:
> > There is a LOT more to it than a simple download in my environment.
> 
> Than things seem to be over complicated in your environment. The test is nothing more than the last dev-build + one downloaded py.

I don't have a range of hardware donated to my development and testing cause.  I only have my two production Beyonwiz T3 units that are merged into an integrated development, test and production environment.  One runs Beyonwiz code and the other current OpenViX Dev code.  There are a number of scripts and processes that manage the current state of these machines. It is a delicate balance that ensures that these machines can also carry out their domestic TV recording duties without compromising the development and/or test code that can also be run on these machines.

original-birdman:
> I hope others will give their opinion on the doubts I reported on saving (see #378 (comment) under A2 and C1) before merging this PR.
...
The sorting is being saved, also to the menu (the latter is new, and I'm not sure if this is desired).

It's being saved to the menu because it's being saved. That's what the original bug report was about - the selected sort method was not being saved.

original-birdman:
> @original-birdman : I think something went wrong, as this is the same file as in the now closed PR #378

Ah!!! You're right!
The line deleted in https://github.com/OpenViX/enigma2/commit/03ebf8a9b5b054ea763252e1dfe147a0aa1cf907 gets _changed_ in this PR, so I'd mised the fact  that I was effectively restorin it.
Thanks for spotting that....

RobvanderDoes:
> > I hope others will give their opinion on the doubts I reported on saving (see #378 (comment) under A2 and C1) before merging this PR.
> > ...
> > The sorting is being saved, also to the menu (the latter is new, and I'm not sure if this is desired).
> 
> It's being saved to the menu because it's being saved. That's what the original bug report was about - the selected sort method was not being saved.
> 

No, that is not correct, and I'm afraid a lot of things are being mixed up now.

Please re-read https://github.com/OpenViX/enigma2/pull/378#issuecomment-463615921 under A2: the sorting was previously already being saved, but not to the menu, and IMHO that behaviour was correct. So now we have an always changing 'default' in per-directory settings which IMHO can't be meant as such. Hence my desire to see others commenting on this before merging.

original-birdman:
> the sorting was previously already being saved, but not to the menu, 

The sorting _wasn't_ being saved.
This whole thing started with a "fix" for:

> In the sortBy() method the newly selected sort mode is not set or saved if the config.movielist.settings_per_directory is True.

except that fix broke sorting for settings_per_directory being false by making all Blue button changes permanent - globally.

original-birdman:
Now on to:
 https://github.com/OpenViX/enigma2/pull/381

RobvanderDoes:
> This whole thing started with a "fix" for:
> 
> > In the sortBy() method the newly selected sort mode is not set or saved if the config.movielist.settings_per_directory is True.

That is not correct; this worked fine and that setting was and is being saved to `.e2settings.pkl`
But you're correct in saying that the setting was _also_ saved to the menu (which IMHO is not desired, but as you said that may be for the next fix).

In reality this whole thing started with a "fix" for:
In the sortBy() method the newly selected sort mode is not set or saved if the config.movielist.settings_per_directory is _False_.

original-birdman:
It is  correct in that the Blue sort key method didn't work. It wasn't doing anything. For the "Sort" setting. (It might have been working for the "SortBy" setting,.)
I checked that before doing anything else, since if it were working there was no need to do anything.

PR #385, opened 2019-03-01
Epg selection glitch

SimonCapewell:
This change adds a selection clip region reset when paging up or down. This fixes a glitch in the EPGList when paging. The steps to reproduce are:
2. Open graph EPG
2. Page down using 8
3. Move up or down one row without making the EPG page. You should get a glitch similar to this
grab. Here I've just moved the selection onto the second row of the second page.

You can also repro going up a page.
You only tend to see the glitch if you're landing on a wider cell after the page. Service name settings make no difference


RobvanderDoes:
There is no way I can reproduce this glitch, so I'll leave committing to somebody who can.

AbuBaniaz:
I could reproduce it.

Huevos:
Happens for me reliably.

IanSav:
I can also reproduce the issue using the steps provided.

PR #386, opened 2019-03-02
Remove invalidate side effect from getTextBoundarySize

SimonCapewell:
`getTextBoundarySize` has a side effect of causing `invalidate` to be called for the parent widget instance. This causes in some cases, a double screen refresh. This has already been worked around in `EpgList.py`, but `TimerList.py` calculates some text sizes per row, so is doing a bunch of invalidates that aren't needed.  Not much of a performance gain - we're talking ~2ms per row of the timerlist to 1ms, but the principal of not causing unintended side effects is a good one.

Something I've not been able to understand is the commonly used pattern of using the heap to allocate locals and managing their lifespan with an `ePtr` allocated on the stack. This appears to just be adding overhead of allocation on the heap (stack is cheaper), and overhead of ref counting! Some examples are in `eSubtitleWidget::event` and `eCanvas::clear`. I've done the static `eLabel::calculate` with no obvious ill effects.


Huevos:
So all "calculateSize" does is return "calculateTextSize"? So why not have all the code in "calculateSize"?

SimonCapewell:
`calculateTextSize` is static, so can be called without instantiating an `eLabel`. The existing method calls onto this static method so that existing callers don't need to be changed.

What's your opinion on the change in TextBoundary.py? It's a bit redundant now that it's easy to call `eLabel.calculateTextSize` instead of `getTextBoundarySize`. Is this likely to be used in any plugins I don't know about?

Huevos:
Not sure if used in plugins. Would be pretty easy to do a search. If you tell me in "quotes" exact text to search I'll test it against all my local gits.

SimonCapewell:
Look for "getTextBoundarySize"

SimonCapewell:
or imports of "TextBoundary"

Huevos:
"getTextBoundarySize"
    enigma2/lib/python/Components/EpgList.py
    enigma2/lib/python/Components/ServiceList.py
    enigma2/lib/python/Components/TimerList.py
    enigma2/lib/python/Tools/TextBoundary.py
    enigma2-plugins/autotimer/src/AutoTimerList.py


"TextBoundary"
    enigma2-plugins/autotimer/src/AutoTimerList.py

PR #390, opened 2019-03-06
Pixmap caching

SimonCapewell:
C++ layer pixmap caching:
`loadPNG` and `loadJPG` have been updated to allow caching using an additional defaulted parameter.
`loadPNG` defaults to cached
`loadJPG` defaults to not cached
Caching is implemented in a helper class in `pixmapcache.cpp`. The cache holds a reference to the pixmap in the same manner as the previous caching. Images are reloaded from disk when the file modified time or the file size changes.
If the cache reaches a size of 256 images, the least recently referenced pixmap is removed. This is, I think, the same behaviour.
`LoadPixmap` has been updated to pass its `cached` parameter through or to use the default rules if it is not supplied by the caller. 

Comments please. There's a fair bit of logging in the caching code currently, this will be removed once everyone is happy with the end result.

I found that the cache caused protection fault crashes on exiting the enigma2 process when the cache held `ePtr<gPixmap>`, so I switched to using a `gPixmap*` and manually managing the refcounts. Since then, no crashes.

### DIFF
--- a/lib/gdi/Makefile.inc
+++ b/lib/gdi/Makefile.inc
@@ -20,6 +20,7 @@ gdi_libenigma_gdi_a_SOURCES = \
 	gdi/lcd.cpp \
 	gdi/picexif.cpp \
 	gdi/picload.cpp \
+	gdi/pixmapcache.cpp \
 	gdi/region.cpp
 
 gdiincludedir = $(pkgincludedir)/lib/gdi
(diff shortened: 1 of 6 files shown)


Huevos:
This has fail Travis ci check.
https://travis-ci.com/OpenViX/enigma2/jobs/182877504

PR #391, opened 2019-03-06
Fix clipping of right aligned service name

SimonCapewell:
When right aligning the service name in the graph EPG, the right edge of the text is clipped slightly short by a few pixels (varies a bit depending on skin). This is also the case if the service number is right aligned and the service name is not being shown.

This change fixes the problem by consistently applying padding around each item when calculating the size of the service rect as well as when building the items to draw.

### DIFF
--- a/lib/python/Components/EpgList.py
+++ b/lib/python/Components/EpgList.py
@@ -96,30 +96,47 @@ class EPGList(GUIComponent):
 			self.l.setBuildFunc(self.buildSimilarEntry)
 		self.epgcache = eEPGCache.getInstance()
 
-		# Common clock icons
-		add = loadPNG(resolveFilename(SCOPE_ACTIVE_SKIN, 'icons/epgclock_add.png'))
-		pre = loadPNG(resolveFilename(SCOPE_ACTIVE_SKIN, 'icons/epgclock_pre.png'))
-		clock = loadPNG(resolveFilename(SCOPE_ACTIVE_SKIN, 'icons/epgclock.png'))
-		zap = loadPNG(resolveFilename(SCOPE_ACTIVE_SKIN, 'icons/epgclock_zap.png'))
-		zaprec = loadPNG(resolveFilename(SCOPE_ACTIVE_SKIN, 'icons/epgclock_zaprec.png'))
-		prepost = loadPNG(resolveFilename(SCOPE_ACTIVE_SKIN, 'icons/epgclock_prepost.png'))
-		post = loadPNG(resolveFilename(SCOPE_ACTIVE_SKIN, 'icons/epgclock_post.png'))
-		self.clocks = [
-			add, pre, clock, prepost, post,
-			add, pre, zap, prepost, post,
-			add, pre, zaprec, prepost, post,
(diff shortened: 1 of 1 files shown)


AbuBaniaz:
With existing file
epg pre simon

With file in pull request
epg post simon

IanSav:
Can the change please be highlighted in the sample images.

AbuBaniaz:
Already highlighted. Have a look at `BBC ONE London`

IanSav:
Thanks for confirming.  My problem is that the "BBC ONE London" text in not right aligned so I didn't see how this correlated to the code change.

My feedback is that the "BBC ONE London" text right padding seems a little excessive.

SimonCapewell:
I'd agree with that. I'll tweak things a little this evening

SimonCapewell:
Actually, the column width turns out to be user defined, so the "right padding" in screenshot 2 is the specified width of the column. What do you think of the padding around the picon and between picon and servicename?

SimonCapewell:
Added another performance improvement. Clocks array was loading the same pixmap repeatedly, which adds up to another 20-30ms load time

Huevos:
Was that an accident adding that to this PR?

SimonCapewell:
Not really. I was nosing around the EPG list and discovered some rather repetetive code. Didn't fancy creating another branch and another PR - this is what's running on my box this evening

Huevos:
Please can you not mix them because otherwise it is impossible to do a "squash and merge" because it means 2 non-related items end up in one commit.

SimonCapewell:
No problem. Another PR incoming

Huevos:
Thanks. If you have a fix to the right hand padding width it can be added here before merging this one.

SimonCapewell:
I don't believe a fix is needed unless the new spacing around the picon is undesirable. The right hand padding as shown in the 2nd screenshot is the width set in Abu's options. I can replicate by setting the servicename width to 200, but if I reduce it to 180, the names are clipped. This is the same sort of behaviour that the previous code had.

Huevos:
Blame Ian, it was him that said it was excessive. ;)

AbuBaniaz:
BTW, I've not altered any options. This is how it looks on vix night hd. Different picon set is being used.

epg post simon_vnhd

IanSav:
> Actually, the column width turns out to be user defined, so the "right padding" in screenshot 2 is the specified width of the column. What do you think of the padding around the picon and between picon and servicename?

That spacing is fine.

(Now you can see why I wanted a pointer at exactly what spacing I was reviewing.  :) )

RobvanderDoes:
Actually I think the spacing around the service number is too wide.

SimonCapewell:
OK, room for some more tweakage. You'll see that the default is `serviceNumberPadding = 9` compared to `serviceNamePadding = 3`, perhaps that 9 needs to be a bit smaller

RobvanderDoes:
Thanks for the tip: I tried `serviceNumberPadding = 3` and that seems fine to me.
But others may find otherwise, depending on skin/font size/number of entries.

PR #396, opened 2019-03-13
Check for fields being False in getInfoCB() rather than just assume t…

original-birdman:
…hat they are strings.

Also re-order some settings and checks.

### DIFF
--- a/lib/python/Screens/About.py
+++ b/lib/python/Screens/About.py
@@ -594,57 +594,84 @@ class SystemNetworkInfo(Screen):
 
 	def getInfoCB(self, data, status):
 		self.LinkState = None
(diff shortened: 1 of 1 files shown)


Huevos:
Currently being tested by Ralf.

Huevos:
Now I can't find the forum thread. Has this been tested yet?

AbuBaniaz:
https://www.world-of-satellite.com/showthread.php?61097-menu-information-network-crash-after-a-few-seconds

IanSav:
The link posted is not open to the public.

Huevos:
No, because our bugs section is not for the public.

PR #400, opened 2019-03-24
H9combo - allow SDcard as mmcblk1p1

TwolDE2:
Like all zgemma boxes, H9combo has different device for SDcard -> mmcblk1p1, so needs different treatment.


TwolDE:
This needs ViX plugin pull as well to work

PR #402, opened 2019-03-28
BUGFIX: Under rare circumstances the timeshift directory can contain …

vampyrebat:
…an .ap file which is not exception handled correctly.  Processing misleads the user into thinking they've incorrectly setup their disk to a format not supporting hardlinks.  There maybe other extensions to exclude or the block of code may want to process other extensions including .ap but in its current state without this patch it will fail and confuse the user.

More details on this thread:

https://www.world-of-satellite.com/showthread.php?61157-Enigma2-Timeshift-Bug


vampyrebat:
Seems to already be incorporated into the Dev branch.

PR #406, opened 2019-04-02
[VirtualKeyBoard.py] Differentiate button icons

IanSav:
Make the tags for icon buttons more obvious by suffixing the tag with "ICON". This also allows the common button names to be available as plain text (translatable) buttons.

The increased use of command icons reduces the need and space complications of translating the command button text.


IanSav:
Remember to grab the equivalent pull request in the skin repository.  ;)

PR #411, opened 2019-04-08
Fix rewind freeze

AbuBaniaz:
this fixes problem described here:
https://forums.openpli.org/topic/67465-rewind-freeze-on-all-dvb-t-mpeg2-recordings/

- Prevent infinite loop(return error if offset didn't change)
- I-frame length calculation changes m_streaminfo -> undo these changes before searching sequence start
- findNextPicture: Start next search after found I-frame
- Adapt comments

### DIFF
--- a/lib/dvb/tstools.cpp
+++ b/lib/dvb/tstools.cpp
@@ -845,8 +845,9 @@ int eDVBTSTools::findFrame(off_t &_offset, size_t &len, int &direction, int fram
 	}
 	off_t start = offset;
 
-	/* let's find the next frame after the given offset */
+	/* calculate length of I-frame -> search next frame -> length = offset - start */
 	unsigned int data;
+	int count_passes = 0;
 	do
 	{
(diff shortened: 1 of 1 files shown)


ViX-Sicilian:
Please hold fire before merging this, need to test a couple of things.

Huevos:
Where did this code come from?

ViX-Sicilian:
OpenATV have this code, not sure where it came from, OpenPLi maybe? Abu able to confirm?

Reason I asked to hold was I though it may be causing an issue with SF8008, but its not. I have no objections, shall we accept into dev?

AbuBaniaz:
As Sicilian was going to run a release build, I did not pick into Dev, instead I created a new branch on my Repo, picked it and then submitted a pull request so it can be merged after the "merge dev so we can run release to incorporate new drivers" process.
Commit is by BetaCentauri on PLi. I didn't hijack it.

Anyway, ill close and we can pick as usual.

PR #419, opened 2019-05-06
Mboot

TwolDE2:
this hopefully resolves all the STARTUP combination names for Zgemma between OpenPli and OE-A and also adds the Image root location for h9 in About


TwolDE2:
can someone try both bootmodes 1/12 on the hd51 and/or h7?? I no longer have the hd51

Huevos:
No HD51 here. H7 not connected right now. Maybe IanSav can test.

Huevos:
I'll start a build off now.

IanSav:
I will be happy to test this out.

Please tell me what I am testing, what should I see and what should I not see.

TwolDE2:
If you have a h7 or hd51 then when you long press the power button and go into the multiboot menu the slot drop down list should give you the choice of rebooting in bootmode 1 or 12 - if you could try flipping between the modes .... after the reboot in menu/information/about it should tell you which bootmode its in. If it doesn‘t work then I need a debug  - we should probably return to posting on teh internal ViX  forum if there are issues

IanSav:
I tried a mode 12 restart from Mode 1 and was greeted with a crash.  I will switch to the forum to discuss the problem.  :)

PR #420, opened 2019-05-09
Ensure timers have only one activation running at a time.

original-birdman:


original-birdman:
For what this fixes see:
   https://www.world-of-satellite.com/showthread.php?57526-Exception-No-such-file-or-directory-tmp-was_rectimer_wakeup

Huevos:
So this is in timer.py. Does it just affect record timers?

original-birdman:
Record timers are the only timers I can see that don't remove themselves from `timer_list` as the first thing they do in `doAcvtivate()`.
_Other_ timers wouldn't be in `timer_list` on the next pass through the loop anyway, so the change is irrelevant to them.

The intention of this loop is to activate timers whose start time has passed. But you only want to do this once per timer.
The code _should_ have been written this way from the outset.

PR #452, opened 2019-08-13
Remove Red and Blue button labels for telnet service as there is noth…

original-birdman:
…ing to do for them.

### DIFF
--- a/lib/python/Screens/NetworkSetup.py
+++ b/lib/python/Screens/NetworkSetup.py
@@ -2351,9 +2351,9 @@ class NetworkTelnet(NSCommon,Screen):
 		self['labstop'] = Label(_("Stopped"))
 		self['labrun'] = Label(_("Running"))
 		self['key_green'] = Label(_("Start"))
-		self['key_red'] = Label(_("Remove Service"))
+		self['key_red'] = Label()
 		self['key_yellow'] = Label(_("Autostart"))
-		self['key_blue'] = Label(_("Show Log"))
+		self['key_blue'] = Label()
 		self.Console = Console()
 		self.my_telnet_active = False
 		self.my_telnet_run = False


Huevos:
I don't understand the commit. If these buttons do nothing (i.e., they are not in the ActionMap) why not remove them. What is the point just removing the text?

original-birdman:
> I don't understand the commit.

So why close it?

> If these buttons do nothing (i.e., they are not in the ActionMap) why not remove them. What is the point just removing the text?

As your own comment says, they are _not in the Action Map_. So how can I remove them from there?
This is removing the text that show up, which implies that there is an action, when there isn't.

Huevos:
Yes but why not remove them completely. You still have the empty label objects there.

Huevos:
BTW, I closed it so et doesn't inadvertently get accepted while being discussed. Re-openning is just a button click.

Huevos:
Also the buttons themselves need removing from any skins that do not have a dynamic button bar.

original-birdman:
> Also the buttons themselves need removing from any skins that do not have a dynamic button bar.

Which would be a good reason to leave them there and just remove the text....

RobvanderDoes:
Naah..... Please don't support lazy skinners.

original-birdman:
> Naah..... Please don't support lazy skinners.

Do you mean that I should just remove the buttons and check that it works with a "standard" skin?

RobvanderDoes:
IMHO not active buttons should be removed; it's up to skinners to dept their skin. We shouldn't stop development because of lazy skinners; skins that are not regularly updated should be removed from our feeds.
A 'standard' screen should be used where ever possible, but that's not related to this issue (but could be done at the same time).

Huevos:
Gordon, I am not bothered skins whose authors don't maintain them. The ability to use a dynamic button bar has been in our enigma code for more than 18 months. I think that is a fair interval to show how little interest these people are showing in their own skins.
https://github.com/OpenViX/enigma2/commit/d1352b35d001bb71f194304a4723c22ee989d6e0#diff-b2a9ece706415994fb9593fe11dfdc8a

Skins with a dynamic button bar show the button image automatically when there is a Label() object created in the .py file. 

In this case the Label() object should be removed as it is not used and then the button text and button image will automatically disappear from all the skins using a dynamic button bar..

All the pre-installed skins have the dynamic button bar.

To get the full function of the dynamic button bar use StaticText() objects, not Label() objects, when creating new code.

original-birdman:
OK. I'll try to get it done (and tested with Simple_1080 and ViXBMC_Confluence) this evening, but if I can't it may not be until next week.

IanSav:
Sorry it has taken me so long to respond to this topic, real life got in the way.

I would also like to see all unsupported skins deprecated.  If no-one is actively supporting a skin then it should go.  It doesn't need to be the original author but someone should be keeping the skins *fully* functional.  I think these deprecated skins should be moved to an archive repository so that the old skins can still be reviewed and potentially adopted by new maintainers.  If the skin is brought back to life it can be restored to the active repository.

If it is of any assistance I have written a guide to managing UI buttons with the new features available: https://github.com/OpenViX/enigma2/blob/master/doc/BUTTONGUIDE

Huevos:
Gordon, Simple 1080 will just work. No need to test that or any pre-installed skin.

In fact don't worry about testing skins at all just remove the Label() objects from the .py and you are done. The button images will still show in skins without dynamic button images. That is up to the skinners to sort out.

A skin like ViXBMC does not have dynamic buttons  In fact it does not have button images at all which is probably why you approach this commit this way.

As a popular skin ViXBMC should be updated to have the dynamic button bar. I will look into this.

Huevos:
BTW, the whole point of the dynamic button bar is so we can add/remove buttons, etc, in .py files and they automatically appear/disappear in the skins without any intervention from the skinners. So for "lazy" skinners that don't want to be  watching lots of repos for changes the dynamic button bar is a must. It also means we can share skin panels between screens that are otherwise identical.

Huevos:
Looking at this further the dynamic button bar is already available in Confluence.

https://github.com/OpenViX/skins/commit/c57b2334e38f10aa343bef788ed8b54593b56d77#diff-0dc918253be7aca7451094dcaf835d1e

original-birdman:
Button setting removed in https://github.com/OpenViX/enigma2/pull/453

Tested on Simple_1080 and ViXBMC_1080_Confluence (since I have a box running each).

PR #455, opened 2019-08-29
Fallback to en_US if chosen language files are unavailable.

original-birdman:


original-birdman:
Look upon this as a temporary hack, to avoid a system crash, until a better solution (make teh chosen files available) can be implemented.

PR #459, opened 2019-09-05
language - change add/menu buttons to add new languages - remove upda…

TwolDE2:
…te cache button


IanSav:
Why was the "Update Cache" functionality removed?

I can see the desire to retrieve the YELLOW button for the "Add Language" function but the "Update Cache" functionality should have been retained under another button.  Perhaps TEXT could be used.

TwolDE2:
Its now done automatically after a change in language and ends up going through the same code loop to finish. The only reason for having the button is to update a table - this now done ... but transparently

PR #464, opened 2019-09-28
Multiboot - simplify Image datestamp code now OpenPli use same source… 

TwolDE2:
code extracted from OpenPli 7.2 rc commit 3bda1e6  (Littlesat) and modified to correctly support Distro name and timestamp for Distro's other than OpenViX now they use same source for timestamp (OpenPli have forgotten other Distro's exist.)

### DIFF
--- a/lib/python/Tools/Multiboot.py
+++ b/lib/python/Tools/Multiboot.py
@@ -103,7 +103,6 @@ class GetImagelist():
 			print "Tools/Multiboot OsPath %s " %self.OsPath
 			if self.OsPath != "NoPath":
 				Creator = open("%s/etc/issue" %self.OsPath).readlines()[-2].capitalize().strip()[:-6].replace("-release", " rel")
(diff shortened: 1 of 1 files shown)


IanSav:
Looking at the bootlogo.mvi as a reference file for image date stamp is unwise.  The logo *can* be changed by users so the date of the file could be effectively random.

TwolDE2:
Don‘t disagree, but its fall back for OpenPLi because of the way they do an automatic restore and pick up time of day, without it they pick up year 1970 ...... OE-A images shouldn‘t go through this part of the code, so no change for OE-A images (OpenViX doesn‘t use this code as we have Rel/Dev versions)

AbuBaniaz:
The initial PLI commit had an error which was then fixed. Does this pull request have that error?

TwolDE2:
@Abu - No I set up the paths to the mounted volume differently
In reality OpenPli now use the same method previously used for other images (except OpenViX) except they have a fallback for when they have year date 1970

PR #465, opened 2019-09-29
PluginBrowser - add Bootlogos

TwolDE2:
This add bootlogos to the list of plugins - I omitted them from list


AbuBaniaz:
Can we discuss this on forum please

PR #480, opened 2019-11-05
Remove any Start/End of Selected Area markers before return from conv…

original-birdman:
…ertDVBUTF8().


original-birdman:
Submitted against real Dev as it seems to work.

AbuBaniaz:
I'd like to change the commit text. I'll merge later tonight

PR #482, opened 2019-11-27
Edision Mio4K /4K+

TwolDE2:
Fix 64bit Arch issues with Buttonsetup and usb backend


TwolDE2:
Note: the ButtonSetup change means that existing Buttons have to be re-applied

AbuBaniaz:
Can we not just cherry pick the commits?

TwolDE2:
The buttonsetup change had conflicts when I cherry-picked into my Git which is why I imported.
The other one feel free to do so - there were no conflicts.

PR #485, opened 2020-01-01
Gigablue add tuners

TwolDE2:


Huevos:
And what about the Vu namings for the same tuner?

TwolDE2:
Problem is that Vu+ and giga use same tuner(s) but different names - don‘t have the Vu+ names

Huevos:
I have added debug to print the names from "_m_description_" variable.

https://github.com/OpenViX/enigma2/commit/596a398285a96b3fa1bab017ce57401b020489eb
https://github.com/OpenViX/enigma2/commit/ff35107677febee575793417abaaa4a8e27b11b0

PR #488, opened 2020-01-23
Fix subtitle runahead on recordings

SimonCapewell:
Some receivers have a bug in the file drivers that deliver subtitles on playback of records completely out of sync with the video, making them unusable. This PR slightly adjusts the bad timing handler to treat subtitles that have been delivered too early by the filesystem as valid and to queue them up for showing later at the appropriate time. The handling of subtitles that arrive too late remains unchanged.

The original fixes for timing issues can be seen in these commits:
- https://github.com/OpenViX/enigma2/commit/f784155
- https://github.com/OpenViX/enigma2/commit/99473c5

If anyone with access to the Travel Channel can test, that'd be appreciated; the first commit states that the changes were added for problems on that channel.

### DIFF
--- a/lib/gui/esubtitle.cpp
+++ b/lib/gui/esubtitle.cpp
@@ -119,7 +119,7 @@ void eSubtitleWidget::setPage(const eDVBTeletextSubtitlePage &p)
 
 void eSubtitleWidget::setPage(const eDVBSubtitlePage &p)
 {
-	eDebug("[eSubtitleWidget] setPage");
+	//eDebug("[eSubtitleWidget] setPage");
 	m_dvb_page = p;
 	invalidate(m_visible_region); // invalidate old visible regions
 	m_visible_region.rects.clear();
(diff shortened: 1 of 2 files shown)


Huevos:
Please do not merge this for the time being. Please send a PR to PLi so we can keep our CPP code in sync as best as possible.

AbuBaniaz:
Surely it makes sense for both teams to test. We don't know what services had issues that warranted reverting the previous fix despite numerous requests. At least if there is an issue we can tweak or revert.

We can also submit a pull request to them if we don't have issues.

Huevos:
Ok, I'll leave you in charges of future PLi merges into these files.

AbuBaniaz:
Closed to keep the peace.

Double standards. Pull request to keep cpp files are closed in some cases. And other fixes are refused citing same converse reason.

AbuBaniaz:
I can confirm this bug on a GB UE 4K.

AbuBaniaz:
Video showing issue and after fix on following link
`
https://mega.nz/#F!B7gVCCBZ!InGsVP5DEBQc1A89NwzEIw
`

SimonCapewell:
Interesting to see that the issue wasn't just affecting the Xtrend machines. It seems crap drivers must be more common than we thought. Thanks for testing on another box.

AbuBaniaz:
@wanwizard tested on PLI. SAB receiver had issue but not Duo 4K. He will test your fix on PLi too.

I have asked that if it is added to PLI, unintended issues in other countries should not result in reverting, we should fix it further.

BTW, I find enabling/disabling subtitles is clunky.

SimonCapewell:
I've got a patch for that. Long press of the subtitles button to toggle the
first subtitles track. Will submit a PR later today

PR #490, opened 2020-01-27
Add quick subtitle toggle feature

SimonCapewell:
This change allows a long press of the subtitles button to toggle subtitles on and off. If there's more than one subtitle track, the first subtitle track is activated. A subtitle style message is shown onscreen when toggling the subtitle to indicate the new status of subtitles - on or off, or if no subtitles are available. The button used is configurable, and on my installation, I'm using a short press for toggle, a long press for the subtitles dialog.

### DIFF
--- a/data/keymap.xml
+++ b/data/keymap.xml
@@ -315,7 +315,8 @@
 	</map>
 
 	<map context="InfobarSubtitleSelectionActions">
-		<key id="KEY_SUBTITLE" mapto="subtitleSelection" flags="m"/>
+		<key id="KEY_SUBTITLE" mapto="subtitleSelection" flags="b"/>
+		<key id="KEY_SUBTITLE" mapto="toggleDefaultSubtitles" flags="l"/>
 	</map>
 
 	<map context="InfobarSubserviceSelectionActions">
@@ -966,6 +967,7 @@
 		<key id="KEY_HELP" mapto="displayHelp" flags="b"/>
 		<key id="KEY_HELP" mapto="displayHelp_long" flags="l"/>
 		<key id="KEY_SUBTITLE" mapto="subtitle" flags="b"/>
+		<key id="KEY_SUBTITLE" mapto="subtitle_long" flags="l"/>
 		<key id="KEY_MENU" mapto="mainMenu" flags="b"/>
 		<key id="KEY_INFO" mapto="info" flags="b"/>
 		<key id="KEY_INFO" mapto="info_long" flags="l"/>
@@ -995,6 +997,7 @@
 		<key id="KEY_PREVIOUSSONG" mapto="rewind" flags="b"/>
 		<key id="KEY_NEXTSONG" mapto="fastforward" flags="b"/>
 		<key id="KEY_SCREEN" mapto="activatePiP" flags="b"/>
+		<key id="KEY_SCREEN" mapto="activatePiP_long" flags="l"/>
 		<key id="KEY_PROGRAM" mapto="timer" flags="b"/>
 		<key id="KEY_PLAYPAUSE" mapto="playpause" flags="m"/>
 		<key id="KEY_LIST" mapto="playlist" flags="b"/>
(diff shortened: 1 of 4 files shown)


AbuBaniaz:
This is causing an issue with the subtitle menu.  Originally, you can access is by pressing subtitle, red button then would be selectable

subtitle_orig_1

subtitle_orig_2

SimonCapewell:
Weird. I've just swapped my short and long press and it's now broken just the same.

AbuBaniaz:
Also, press long subtitles to activate/deactivate. Ideally that should be end of it. However, you then get into menu. I hope that makes sense.

SimonCapewell:
Actually, I can't repro the red button problem. I just didn't realise that it's not available when subtitles are off.

The menu popping up when letting go of subtitle, I can reproduce though.

AbuBaniaz:
>I just didn't realise that it's not available when subtitles are off.
Makes two of us. Makes sense as as you can't alter position/colour etc if no subtitles are showing.

Most skins do not have the menu button either.

SimonCapewell:
I think fixing the rogue popup on long button press is going to be a bit tricky. I can't see an obvious way to prevent the bounce event when you release the key.

SimonCapewell:
This should fix it. I had to introduce a global key watcher to do though. The e2 action map and keypress handler isn't really geared towards dealing with this long button press scenario.

AbuBaniaz:
Many thanks. Works very well.

AbuBaniaz:
@SimonCapewell 
There is a bug in this. We can't disable the buttons presses when a message is showing. 

Try doing the following:
Backup your lamedb first.
Remove all services and restart. (up next to OK, menu, highlight terrestrial services, menu, delete, restart.

You should come up with a service unavailable message. You can't get out of the screen.

We'll have to revert before next build unless fixed.

AbuBaniaz:
A release needs to be run. As we always merge dev to release, I have been asked to revert this.

SimonCapewell:
@AbuBaniaz I can't see any option to remove all services using the steps you listed. So, I've removed lamedb and bouquet files, but still no service unavailble popup on startup. Triggering the popup on e2 startup in code doesn't show any ill effect. Can you have another go testing it, but remove the `eActionMap.getInstance().bindAction('', -maxint - 1, self.keypress)` from SubtitleDisplay.py. I can't see at the moment how the new code can be causing this problem.

AbuBaniaz:
I'll try the new code, problem is I am stuck on PLi owing to the multiboot code they are changing. I'll send @TwolDE2  a PM to see if he can run it too.

Just for reference and in case it makes things clearer.
Press Up next to OK. Then Green






Alternatively, init 4, wait for 10 seconds, rename lamedb, init 3

SimonCapewell:
Thanks. Looks like I did the terminal steps, so there's probably some settings that are different

original-birdman:
Abu asked me to test this.

I downloaded the code in this change (as a patch) then:

- Installed patches and restarted GUI.
    Long-subtitle toggles OK.
    Short-subtitle brings up menu.

- Removed terrestrial services by:
     Up next to OK
     Green (Extended - but I did get it to show Satellites once....)
     Highlight "Terrestrial Services"
     Menu
     Remove terrestrial services
     Restart GUI or reboot
 
- Restarted GUI.
     Timer problem pop-up. Cancelled it.
     Service Unavailable. Cancelled it.
     Can bring up menus.
 
- Rebooted box.
     Service Unavailable. Cancelled it.
     Can bring up menus.

So, for me, the code here seems to be OK as-is.

AbuBaniaz:
I cannot reproduce the issue again. I have re-added this having consulted Stef. He will run a Dev build tonight.

SimonCapewell:
There's a bit of crappy code in there around removing the keybinding on close (or in fact not removing it) but this'd only be encountered when shutting down, and it won't matter then. I can submit a tidy up at a later date.

AbuBaniaz:
PLI have asked for a pull request to be submitted for this. So I'll wait for your modified code.

SimonCapewell:
https://github.com/OpenViX/enigma2/pull/493

PR #492, opened 2020-01-30
Harddisk

TwolDE2:


TwolDE2:
This pull moves the filesystem check for Udev/Devfs to SystemInfo and then uses the SystemInfo["Udev"] variable in Harddisk.
Also Harddisk imports are sorted and a little code cleanup

PR #498, opened 2020-02-17
Zap picons

SimonCapewell:
This PR allows skins to display a picon element in the NumberZap screen by adding a ServiceEvent object named `Service`.

The example below is a number zap in the top left of the screen the number with the picon to the right:
```
	<screen name="NumberZap" position="20,20" size="330,100" backgroundColor="outlinemenubackground" flags="wfNoBorder">
		<eLabel position="175,5" size="150,90" halign="center" backgroundColor="menubackground" zPosition="-1"/>   
		<widget name="number" position="5,5" size="165,75" font="menulist;80" foregroundColor="maincolour" backgroundColor="menubackground" transparent="1" borderColor="menubackground" borderWidth="3" valign="center" halign="center" zPosition="2" />
		<widget source="Service" render="Picon" position="180,10" size="140,80" alphatest="blend" backgroundColor="background" transparent="1">
			<convert type="ServiceName">Reference</convert>
		</widget>
	</screen>
```


AbuBaniaz:
I have picked in the picon refresh bug fix. I think there is a bug thread for it.

Can you make the picon option configurable please? I attach a picture to show effect that it has
numberzap picon

SimonCapewell:
Best wait until I've worked out why the LCD picon doesn't load up on bootup before proceeding further.

Configurable picon on the numberzap is do-able, but how does this work with skins? If I've got a background large enough for a picon (Confluence for example), it'll be a big blank space when the picon option is turned off, unless there's an autosize ability

AbuBaniaz:
Just passing on a request to cater for people who may not want them. I don't think we have an option for infobar either, but that is another topic.

SimonCapewell:
Yeah, NP adding the option, just something to be aware of. Not an issue for the skins that have transparent numberzap

SimonCapewell:
I've added the ability to use a NumberZapPicon screen, same as ATV have, so there's some limited support already out there in Metrix and Kiddac skins.

PR #501, opened 2020-02-22
Add debug output to listbox python buildfunc calls

SimonCapewell:
This change adds exception stack output
- when a Python buildfunc throws an exception
- when a button press that opens a screen throws an exception


AbuBaniaz:
A Release build is imminent, so this will be merged after that

PR #514, opened 2020-03-28
Provide option to hide done timers

SimonCapewell:
Updates to timer edit list:
- Add another option to hide finished timers
- Add ch+ and ch- to move to top and bottom of timer list, similar to movie list
- Removal of commented out debug print lines

### DIFF
--- a/data/keymap.xml
+++ b/data/keymap.xml
@@ -688,6 +688,8 @@
 	<map context="TimerEditActions">
 		<key id="KEY_INFO" mapto="log" flags="m"/>
 		<key id="KEY_I" mapto="log" flags="m"/>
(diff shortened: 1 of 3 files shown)


AbuBaniaz:
Where is the option to hide timers meant to be configured?

Just for info, old timers were usually kept to avoid duplicate recordings.

SimonCapewell:
It's the 12th item on _User Interface | Settings_. _Position of finished timers_ 

Yes, it's fine to keep the timers to avoid duplicates, just not very useful to have them clogging up the list!

SimonCapewell:
You're partly right. I should've included UsageConfig.py

Huevos:
Is this ready to merge?

AbuBaniaz:
Done.

PR #515, opened 2020-03-29
Added project information to README.md

stappersg:
Tells something more about what the  enigma2 project is about ...


Huevos:
How do you think this is a relative of OpenWRT?

PR #536, opened 2020-05-12
Allow length of Service Name to be specified

bbbuk:
Allows length of service name within MultiEPG to be specified depending on layout of skin.  Min of 5, max of 20 and defaults to 7 (matching existing hard-coded length).

This option only applies to MultiEPG because by default, depending on skin, service name may be cut off.  The length of this can't be easily set as it depends on skin in use so this option allows user to set this themselves.

Thanks to @seagen for highlighting this issue. More details mentioned in [this](https://www.world-of-satellite.com/showthread.php?62653-New-EPG-files-User-trial-and-feedback&p=497546&viewfull=1#post497546) thread.

### DIFF
--- a/data/setup.xml
+++ b/data/setup.xml
@@ -320,6 +320,7 @@
 		<item level="2" text="OK button (long)" description="Set to what you want the button to do.">config.epgselection.multi.btn_oklong</item>
 		<item level="2" text="Number of rows" description="Configure the number of rows shown.">config.epgselection.multi.itemsperpage</item>
 		<item level="2" text="Event font size" description="Configure the font size relative to skin size, so 1 increases by 1 point size, and -1 decreases by 1 point size.">config.epgselection.multi.eventfs</item>
+		<item level="2" text="Service Name Length" description="Configure the length of ServiceName">config.epgselection.multi.servicename_length</item>
 	</setup>
 	<setup key="epgsingle" title="Single EPG">
 		<item level="2" text="Channel selection" description="Choose what channel to select when the EPG opens.">config.epgselection.single.browse_mode</item>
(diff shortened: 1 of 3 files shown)


bbbuk:
Sorry looks like I accidently chose master branch but I did fork off dev branch just yesterday and created my own branch (thanks to Abu for instructions) so this shouldn't cause a conflict.  Has been tested by several users including myself.

IanSav:
Is this the appropriate way to fix the issue?  Given that the font sizes are adjustable varying the number of characters would seem to be the wrong way to restore columnar alignments.

I suggest a better way to fix this would be to add skin parameters that define the widths of the columns to ensure that no matter what the skin the columns can be made to align.

bbbuk:
If you're referring to automatically adjusting column based on service name then that sounds okay except if you have a long service name, it would mean column being automatically adjusted to fit that which would mean that you may lose some of the now/next information.

At least if this is user preference, they could choose what they can do with/without so if they decided service name length to be long to accommodate a long servicename then they will lose some information to right of it like now/next epg info especially if you have PiP as I do on my skin.

However, equally important maybe to have just right length of service name so can at least work out what channel is called but then have the more space for the EPG now/next.

I'm open to suggestions?

AbuBaniaz:
Thanks
I have committed slightly modified files, based on your work. I used names/text that already exists .
https://github.com/OpenViX/enigma2/commit/2029d3c9f858fb9a29967b15c82de2ae7fcfc7c9

IanSav:
I was proposing that there could be a parameter with three width settings one width for the service column, one for the progress or event times and one for the event name.  Once the column widths are defined then the text in each column will naturally be truncated when the text string exceeds the allocated column width.  There is no need to set or play with the actual string lengths themselves.

PR #543, opened 2020-05-16
Detect split programs when setting timers

SimonCapewell:
Many commercial channels split longer programmes, usually films, with short news bulletins. This change adds an option to detect this and automatically extend the recording across the news so you don't accidentally miss the second half. Most places in the UI support now this:
- 2nd infobar
- Channel List
- EPGs
- Event View

A new setting, `split_programme_minutes` has been added, which defaults to 15 minutes, so should cover both the entertainment style news and the longer bulletins seen on ITV from time to time. We can reduce the default or have it switched off by default if necessary.

The detection is uses the name _and_ short description of the events. This seems to work well on the UK's ITV, 5, Sony and Paramount channels.

### DIFF
--- a/RecordTimer.py
+++ b/RecordTimer.py
@@ -36,21 +36,35 @@ from sys import maxint
 import threading
 write_lock = threading.Lock()
 
-# parses an event, and gives out a (begin, end, name, duration, eit)-tuple.
-# begin and end will be corrected
-def parseEvent(ev, description = True):
+# Parses an event, and returns a (begin, end, name, duration, eit)-tuple.
+# begin and end include padding (if set in config)
+# If service is supplied, end will also include any split program spanning adjustment (if set in config)
+def parseEvent(event, description=True, service=None):
 	if description:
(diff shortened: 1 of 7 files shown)


RecordTimer.py
```diff
@@ -1298,9 +1322,7 @@ def isInTimer(self, service, begin, duration):
 		end = begin + duration
 		startAt = begin - config.recording.margin_before.value * 60
 		endAt = end + config.recording.margin_after.value * 60
-		if isinstance(service, eServiceReference):
```
SimonCapewell, line 1301: This was causing a crash due to the bad design of the `ServiceReference` class.

AbuBaniaz:
The benefits of this commit are as pictured. "About a boy is in two segments" between 21:00 and 23:05. The "Access" event is sandwiched in between at 22:00 for five minutes.

EPG
epg

Without the commit, the timer is set only for one segment
without_split_timer_commit

With the commit, the timer is set for the whole movie
with_split_timer_commit

PR #550, opened 2020-05-22
[ButtonSetup] Add Long press Tv and Radio buttons and the option to r…

Ev0-BH:
…un a script from /usr/script

(cherry picked from commit 82e52c77a98c7af77ca52166ae732123e29e3552)

### DIFF
--- a/data/keymap.xml
+++ b/data/keymap.xml
@@ -978,7 +978,9 @@
 		<key id="KEY_RECORD" mapto="rec" flags="b"/>
 		<key id="KEY_RECORD" mapto="pvr_long" flags="l"/>
 		<key id="KEY_RADIO" mapto="radio" flags="b"/>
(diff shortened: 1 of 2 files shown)


TwolDE2:
Surely you can run scripts from ViX plugin, so why would you do this hack????

Huevos:
> 
> 
> Surely you can run scripts from ViX plugin, so why would you do this hack????

Yes you can. No idea why this would be needed.

Ev0-BH:
So you can set a script to run from a single button press.

TwolDE2:
But you can do that already from the ViX plugin - this doesn‘t add any value other than code for a personal wish

Ev0-BH:
Well OBH doesn't have the vix plugin, which is why I added it.

If you don't want / need the code, feel free to close the request, it's not a problem :)

PR #559, opened 2020-06-04
Add template element.

SimonCapewell:
Templates can be referred to by panels in the same was as screens
but are positioned relatively, making it easier to reuse elements on screens with
slightly different dimensions


Huevos:
Closed, until the way forward is fully discussed and agreed in the forum thread. Thanks.

PR #561, opened 2020-06-21
Change primetime to left edge

AbuBaniaz:
Change primetime to set the time of the left hand edge of the grid EPG

### DIFF
--- a/lib/python/Components/EpgListGrid.py
+++ b/lib/python/Components/EpgListGrid.py
@@ -182,9 +182,11 @@ class EPGListGrid(EPGListBase):
 			self.skinAttributes = attribs
 		return EPGListBase.applySkin(self, desktop, screen)
 
-	def setTimeFocus(self, timeFocus):
+	def setTimeFocus(self, timeFocus, center=True):
 		# prefer time being aligned in the middle of the EPG, but clip to the maximum EPG data history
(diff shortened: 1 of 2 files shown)


Huevos:
The primetime should be shown allowing for the amount of history that the user ASKED to be displayed!

Randomly jumping to some point against the wishes of the user is not a good approach.

AbuBaniaz:
I have already cherry picked this bug fix. Simone had added it to his epg custom button branch. 

Thread here:
https://www.world-of-satellite.com/showthread.php?62860-Wrong-Primetime

I had submitted a pull request so it is in the queue of commits to be added while we await migration to new branch. As other commits have started going in, I added it.

Huevos:
Sorry, not sure what you mean about a queue but this commit can't be used.

AbuBaniaz:
No problem. Feel free to revert the commit and leave bug in place.

SimonCapewell:
Happy for this to be reverted, but I'm out of ideas now regarding the bug report about primetime not behaving as expected.

Huevos:
This is like the multi-bouquets bug. First we need a clear list of steps to reproduce the bug. Without that we don't have a lot of hope except finding by brute force. Could it be something to do with program length?

PR #566, opened 2020-07-01
Use ImportError to detect missing plugins

SimonCapewell:
Switch InfobarGenerics to use _ImportError_ as the indication of a missing plugin rather than relying on the presence of _plugin.pyo_


Huevos:
Simon, I didn't see this PR. We seem to have tripped over each other. Please check you are happy with my commit. I only have the import in the try block as that is what we are testing.

https://github.com/OpenViX/enigma2/commit/d34a85ca0c9e430ce9116418441ce2ba349532af

PR #573, opened 2020-08-19
Disable native mappings

prl001:
Patched from working code on the Beyonwiz image.

One patch hunk did failed because of other minor differences between the two image repository states and was edited by hand. The edits were simple, and hopefully should compile cleanly and run.


prl001:
PR to wrong repo & branch!

PR #574, opened 2020-08-20
Fix Menu to stop crashes:

Ev0-BH:
Remove Common Interface from Menu for boxes with no CI slot.
Fix crash on boxes with no front panel display


Huevos:
Why has this been closed?

PR #580, opened 2020-09-30
Mask any Wifi password from showing in the debug log.

original-birdman:
Replaces any "-k paswword" with "-k XXXX".

### DIFF
--- a/lib/python/Components/Network.py
+++ b/lib/python/Components/Network.py
@@ -221,7 +221,15 @@ class Network:
 			# load ns only once
 			self.loadNameserverConfig()
 			print "[Network] read configured interface:", ifaces
-			print "[Network] self.ifaces after loading:", self.ifaces
+# Take a copy of self.ifaces and mask any Wifi password in it...
+# The only password will be a preup -k option to wl-config.sh, set in
+# lib/python/Plugins/SystemPlugins/WirelessLan/plugin.py
+#
+			safe_ifaces = self.ifaces.copy()
+			for intf in safe_ifaces:
+				if 'preup' in safe_ifaces[intf] and safe_ifaces[intf]['preup'] is not False:
+					safe_ifaces[intf]['preup'] = re.sub(' -k \S* ', ' -k XXXX ', safe_ifaces[intf]['preup'])
+			print "[Network] self.ifaces after loading:", safe_ifaces
 			self.config_ready = True
 			self.msgPlugins()
 			if callback is not None:


Huevos:
So if we hide part of the debug how are we supposed to read it?

RobvanderDoes:
I fail to see any reason to accept this PR. Nobody has access to my logs, unless I decide so myself.
And how could I chech if XXXX coincides with my password?

ViX-Sicilian:
I think the main use for this would be if someone wanted to submit a debug log on the forum their debug key would be removed from the logs.

RobvanderDoes:
Even so: what could I possibly do with the knowledge of a password of "vix-forum-user-xxxx"?

ViX-Sicilian:
> Even so: what could I possibly do with the knowledge of a password of "vix-forum-user-xxxx"?

You never know what passwords some people use. I'd never use anything sensitive, but you never know what some do.

RobvanderDoes:
And how could I check if XXXX coincides with my password?

ViX-Sicilian:
> And how could I check if XXXX coincides with my password?

Why would you want to? If you want to check only the password is being masked, check before and after logs.

SimonCapewell:
Surely you check the password by the fact the box isn't connecting? Storing passwords in logs is a bad idea.

RobvanderDoes:
No, It's the other way around: if the box doesn't connect, I need to find out why that's the case.

RobvanderDoes:
Why would you want to? If you want to check only the password is being masked, check before and after logs.

How could I possibly check if the password is not logged?

ccs-ccs:
.....  Some boxes have never logged the wifi password, some do, does yours?

RobvanderDoes:
So some boxes run a different E2?

SimonCapewell:
I was about to go off and check my logs. Then I remembered it's wired :laughing:

ccs-ccs:
**So some boxes run a different E2?**

No.

Same ViX, same room, same passord, one does, one doesn't.

As a poor old end-user, I was most unimpressed to see my wifi passord now in the public domain.

RobvanderDoes:
Same ViX, same room, same passord, one does, one doesn't.

So one needs to keep the password logged to find out why, as this looks like e bug.

ccs-ccs:
> Same ViX, same room, same password, one does, one doesn't.
> 
> So one needs to keep the password logged to find out why, as this looks like a bug.

I doubt it, more likely because one tuner is built in, the other is usb.

original-birdman:
Whether the password is in the logs depends in which sub-system is running the Wifi.
There's MadWifi and "the rest".
One runs wpa-supplicant (which stuffs the password elsewhere) and the others uses wl-config.sh.

Note that wpa-supplicant set-ups don't show the password in the debug log _anyway_, so the argument about "how will we tell from the debug log what password is being used" has no weight, since for many adaptors you've **never** known in the past anyway.

ccs-ccs:
OpenPLi seem quite keen on removing passwords.......

https://github.com/OpenPLi/enigma2/commit/dedc7d4aafb545000d94b4a0cc8290b6eea8e9b1

original-birdman:
The code to `print "read configured interface:"` and `print "self.ifaces after loading:"` _used_ to be commented out.
It looks as though it was uncommented for testing during a set of code changes and not re-commented.

https://github.com/OpenViX/enigma2/commit/ccd9e63e2044d69f0d03964f5a6fd807f9829cef

ccs-ccs:
Print statements commented out in 2012 ......

77d19dfb303e78f74679dea89c15f03abde20c5d

Huevos:
I really can't see a security issue here, unless some fool uses their credit card number as a WPA2 key.

Also the password is easily accessible to anyone who has access to the box (not from the logs).

Huevos:
And, by the way this is not a password, it is a WiFI access key, for a WiFi network that has a range of about 50 feet.

original-birdman:
```
by the way this is not a password, it is a WiFI access key
```

I'll remember that the next time I go into a pub that displays its Wifi password and refuse to use it, instead demanding they give me the Wifi access key.

ccs-ccs:
> I really can't see a security issue here, unless some fool uses their credit card number as a WPA2 key.
> 
> Also the password is easily accessible to anyone who has access to the box (not from the logs).

So what's the reason for **not** suppressing wifi passwords?

If access to the box is restricted to the owner, then wifi passwords are **not** published in the public domain, unless, of course, they are fool enough to help debugging problems by providing debug logs.

Maybe I should open a thread on the ViX forum and see what the reaction is? How would you like me to word it?

Huevos:
This is in debug logs. The whole point of debug logs is to contain debug info. This PR is making a mountain out of a molehill.

ccs-ccs:
> This is in debug logs. The whole point of debug logs is to contain debug info. This PR is making a mountain out of a molehill.

Since when have people's private passwords been a necessary part of debugs logs?

Huevos:
It is a WiFi key and this is not the olny file that contains it. Please explain how you think this is a security risk.

ccs-ccs:
> It is a WiFi key and this is not the olny file that contains it. Please explain how you think this is a security risk.

It's the only file that gets regularly published on a public forum.

Is there a security risk? I reckon the average punter probably believes there is, and that's all that really matters.

I don't really understand why such a simple PR has been greeted with such derision. I'm amazed, to be honest.

That's me done, clearly you'll reject it.

BrianG61UK:
> 
> 
> It is a WiFi key and this is not the only file that contains it. Please explain how you think this is a security risk.

Okay. I demand you tell me your Wi-Fi "password" NOW or stop saying nonsense like it's not a security risk.

RobvanderDoes:
No problem, here's mine: €7Wk9;%gEP4

ccs-ccs:
> No problem, here's mine: €7Wk9;%gEP4

Maybe you could also include your latitude/longitude or post code, they often get posted as well?

RobvanderDoes:
Is my postcode in the debuglogs then?

ccs-ccs:
> Is my postcode in the debuglogs then?

... "often get posted" is the clue.

RobvanderDoes:
I arrest my case; a paranoid man can (by definition) never be convinced.

PR #586, opened 2020-10-29
Fix Channel List record menu popup crash

SimonCapewell:
Fixes a crash introduced in 5.4.001


SimonCapewell:
master is broken, but Dev is already fixed.

PR #590, opened 2020-11-22
Fix recording deadlock

SimonCapewell:
Fixes for a threading deadlock when a recording is finalised and the recording thread is requested to be shut down.


SimonCapewell:
I'll put a bit more detail on the next PR :-)

PR #602, opened 2020-12-07
HdmiCec.py handle use one instance in py2/py3

TwolDE2:


HdmiCECInstance.py
```diff
@@ -0,0 +1,2 @@
+
+instance = None
```
SimonCapewell, line 2: Why not just put this in HdmiCec? We have a similar pattern for the infobars

TwolDE2: I am open to whatever works - I just used this method as its the one used in Navigation(which I know code wise, just never looked at Infobars))  and the method works for python2 & python 3

TwolDE2: I will kill the Pull and try this in Py3 (no reason it shouldn't work)

PR #605, opened 2020-12-10
Fix small bugs

SimonCapewell:
Fix a raft of minor irritations. Details in each commit.

### DIFF
--- a/Navigation.py
+++ b/Navigation.py
@@ -200,17 +200,8 @@ class Navigation:
 	def getCurrentlyPlayingServiceOrGroup(self):
 		return self.currentlyPlayingServiceOrGroup
 
-	def isMovieplayerActive(self):
-		MoviePlayerInstance = MoviePlayer.instance
-		if MoviePlayerInstance is not None and '0:0:0:0:0:0:0:0:0' in self.currentlyPlayingServiceReference.toString():
-			from Screens.InfoBarGenerics import setResumePoint
-			setResumePoint(MoviePlayer.instance.session)
-			MoviePlayerInstance.close()
-
 	def recordService(self, ref, simulate=False):
 		service = None
-		if isinstance(ref, ServiceReference.ServiceReference):
-			ref = ref.ref
 		if not simulate:
 			print "[Navigation] recording service:", (ref and ref.toString())
 		if ref:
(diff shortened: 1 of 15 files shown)


RobvanderDoes:
Sorry to say Simon, but normally it's a good idea to have just one fix per commit.

Huevos:
> 
> 
> Sorry to say Simon, but normally it's a good idea to have just one fix per PR.

It's fine if the person accepting the PR does a rebase... which hasn't happened in this case... so now there is junk in the commit history.

RobvanderDoes:
Not sure why this has been accepted.
The normal way is yo have one commit per PR; that way each individual commit can be tested and accepted/rejected.
We always do it that way, so what's the reason to accept a bulk PR now?

RobvanderDoes:
Should we revert and do it the proper way, or do we leave it as is?

IanSav:
> Not sure why this has been accepted.
> The normal way is yo have one commit per PR; that way each individual commit can be tested and accepted/rejected.
> We always do it that way, so what's the reason to accept a bulk PR now?

Good question.  I was always made to resubmit all contributions that were combined or had messy commit histories.

RobvanderDoes:
Also reverting just one of the commits, should that be required, is now quite difficult.

Huevos:
> 
> 
> Also reverting just one of the commits, should that be required, is now quite difficult.

Why? You would just revert the individual commit.

AbuBaniaz:
When I previously used to ask other people to submit one pull request per issue, I was jumped on. And the multiple commits were merged as one commit. I took that as being fine now.

Now that  I did the same but accept multiple **individual** commits, you say the opposite. The commits (apart from merge commit) are separate and you can revert them individually. You can always view history without merge commits. Anyway, I'll get rid of the bad history.

AbuBaniaz:
I have removed the merge commit locally, but don't have rights to push the changes to GitHub.

Huevos:
Squashing down to one commit means individual commits are combined into an unreadable mess.

Doing a rebase means all individual commits show.

What we don't want/never wanted is the pointless merge commit.

AbuBaniaz:
> Doing a rebase means all individual commits show.
 They do too on a merge commit. 

> What we don't want/never wanted is the pointless merge commit.
Until Rossi started to use the squash and merge option. We used to do that anyway.

To get rid of the merge commit I added but fine for what others add, see screenshot below. As I said, I don't have the rights to do it.
branches

Huevos:
Removing a commit afterwards would break everyone's local and the build.

PR #622, opened 2021-01-20
Fix ParentalControlSetup pin popup

zukon:


AbuBaniaz:
Thanks. I have added the commit to Dev branch, so I will close this pull request. It will be merged into Release branch when the next image is built. We have just missed the one that is running.

PR #633, opened 2021-02-10
Master

jeepcook:
Hello,

Concerned github repository enigma2/data.

When the groupedservices file is correctly set, if you select a channel list in, you can display all sub channels defined in the groupedservices file by pressing green button.

But the only displayed information is the current epg for every sub channel. So if you have the same sub channel but with different versions, we are unable to know which one to select, because they have the same now epg, and no name. Same thing if you have a lot of distinct channels, because the channel name is not displayed you are unable to choice (example with the HD+ bouquet set in the default groupedservices file).

I think the solution could be to display "channel Name | now epg". For the channel name, the better solution could be to display the name of the channel set in the groupedservices files:

Example: for 2 french channels, their names are the same in lamedb (One is TF1, the second one FRANCE 2), so displaying the channel name extract from lamedb is not a solution because they are the same. But displaying the channel name set after the # could be great.
TF1

1:0:19:2460:440:1:C00000:0:0:0: #TF1 CSAT
1:0:19:601:6:20FA:DDE0000:0:0:0: #TF1 PLS TNT
1:0:19:191:4FB0:55F:DDE0000:0:0:0: #TF1 FRANSAT
FRANCE 2

1:0:19:245A:440:1:C00000:0:0:0: #FRANCE 2 CSAT
1:0:1:101:1:20FA:DDE0000:0:0:0: #FRANCE 2 PLS TNT
1:0:19:12D:4F4C:55F:DDE0000:0:0:0: #FRANCE 2 FRANSAT
.......

If you need some other informations, please contact me.

Thanks


AbuBaniaz:
Please send your pull request to our develop branch

jeepcook:
Hello, finally I opened an issue because in the current mode this functionality is unusable.
Thanks

PR #646, opened 2021-02-24
Optionally group recordings by name into collections

SimonCapewell:
Collections are virtual folders, similar to those seen in the Dream EPG
Android app. They automatically group episodes of a programme together
so you don't need to create directories yourself.
The movie list displays and navigates them as if they're directories.
Delete, move and copy on collection acts on all files in the collection
There's also an option to not group if items are already organised in a
same named directory.


SimonCapewell:
There's a new icon required in the default skin for this. PR later

PR #649, opened 2021-03-02
Fix missing translation strings

SimonCapewell:


SimonCapewell:
Looks like this has been done already 👍

Huevos:
Yes, sorry, I was fixing this other translation bug which was a bit of a nightmare.

PR #656, opened 2021-03-11
[Satfinder] Remove duplicated entry.

nunigaia:
Line 171.

Removed the word "Tuner" from line:

		self.satfinderTunerEntry = getConfigListEntry(_("Tuner"), self.satfinder_scan_nims)

This avoids the duplicated entry of the word tuner at SKIN satfinder Panel.


Huevos:
If there is an issue in your skin fix it in the skin.

PR #665, opened 2021-03-24
Add a toDisplayString() to the ConfigClock and ConfigDateTime classes.

original-birdman:


original-birdman:
This means that, e.g., the help text when setting a new timer says "24 March 2021" (not "1616586120") and "11:42" (not "[11, 42]" for the default settings.

PR #667, opened 2021-03-25
[SoftCam] Add Support For Ncam

fairbird:


Huevos:
A team member needs to test this before accepting it.

AbuBaniaz:
I am getting lockups with an endless spinner this when starting from the GUI. Even for  Oscam-emu. Debug logs don't show anything.

fairbird:
I will create new pull request

PR #692, opened 2021-05-18
Fix movielist blank rows

SimonCapewell:
Also use words in the timer log for the timer activation state


Huevos:
Is this for ready for merging? I didn't follow the comments about hasattr.

SimonCapewell:
Hold off for now. I'd mark it as draft again if I could see a button to do so

PR #698, opened 2021-07-08
PEP8.sh: Use date instead of Travis build number as nobody uses Travi…

persianpros:
…s anymore and we run this script monthly (usually/manually)


Huevos:
Thank you.

PR #705, opened 2021-08-22
Enable log rotation to prevent log corruption

original-birdman:


original-birdman:
The current TrimJob code attempts to move data from the end of a file to the beginning; discarding the beginning. This corrupts any current log, as it is the start-up shell that is writing to it and it continues to write at *its* end of file.

Since it is the start-up shell script that "owns" the log, it has to rotate it.

This adds handler for when a file reaches a size limit.
Changes to the log size and its location are just echoed and will be acted on by the shell.
Log files now state the previous (at the start) and next (at the end) log file name.

This is the Python3 version.

Huevos:
The formatting has been changed in enigma2.sh. This needs to be returned to the original.

Huevos:
Closed for the time being so it is not merged until ready.

original-birdman:
> The formatting has been changed in enigma2.sh. This needs to be returned to the original.

Why?
There is no need to have such indentation in a shell script.

PR #715, opened 2021-09-29
Only updateEventInfo if there is a current selection.

original-birdman:


original-birdman:
I had a crash here - possibly when no movie location was set(?).

PR #718, opened 2021-11-05
Ensure that the logger exits when enigma2 does.

original-birdman:


original-birdman:
This solves the issue of enigma2 firing off processes (such as bsa_server) which keep stdout/stderr open, preventing the logger exiting via a closed input stream.

PR #722, opened 2021-11-06
Dev python3 compatible Remove control characters from EPG text

BrianG61UK:


BrianG61UK:
Dev python3 compatible Remove control characters from EPG text
(at end of function convertDVBUTF8 in lib/base/estring.cpp)
See https://www.world-of-satellite.com/showthread.php?64435&p=518234&viewfull=1#post518234 and previous discussion.

PR #725, opened 2021-11-25
Use str not int for fallback setting, otherwise find() will fail.

original-birdman:


original-birdman:
No example of it failing. I was just checking code for `x and x.func()` type settings and spotted this.

PR #726, opened 2021-11-27
[ChannelSelection] Fix crash error with some plugin...

fairbird:
<  4221.2589> 19:55:56.1865   File "/usr/lib/enigma2/python/Screens/ChannelSelection.py", line 2043, in __init__
<  4221.2611> 19:55:56.1887     assert ChannelSelection.instance is None, "class InfoBar is a singleton class and just one instance of this class is allowed!"
<  4221.2613> 19:55:56.1888 AssertionError: class InfoBar is a singleton class and just one instance of this class is allowed!
<  4221.2614> 19:55:56.1889 [ePyObject] (CallObject(<bound method ActionMap.action of <Components.ActionMap.ActionMap object at 0xb33f6658>>,('GlobalActions', 'showRaedQuickSignal')) failed)

### DIFF
--- a/lib/python/Screens/ChannelSelection.py
+++ b/lib/python/Screens/ChannelSelection.py
@@ -2039,9 +2039,12 @@ class ChannelSelection(ChannelSelectionEdit, ChannelSelectionBase, ChannelSelect
 				iPlayableService.evEnd: self.__evServiceEnd
 			})
 
-		if type(self) is ChannelSelection:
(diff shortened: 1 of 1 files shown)


Huevos:
That assert is there for a reason. ChannelSelection is a singleton so no code should be attempting to open additional instances.

I have already provided a fix for the plugin that is attempting to do this.

PR #729, opened 2021-12-06
Force a reload if a name-change changes a meta file.

original-birdman:


RobvanderDoes:
Should this still be applied after openvix: developer 6.0.004.011 has been build?

original-birdman:
It needs to be applied at some point, but it could wait for 6.0.005.001.
It's not crucial - it's not fixing a crash.

PR #734, opened 2021-12-19
New finnish (fi.po) translation.

Orlandoxx:


Huevos:
Hi, still can't be rebased. I will fix it. You can see there is a problem because your PR contains non relate commits.

Huevos:
Sorry I have had to revert this. Looks like your repo is broken. Easiest thing would be delete it and refork. And work in a feature branch not the main branch. And create a new feature branch for every new PR.

PR #744, opened 2022-01-08
Updated finnish translation (fi.po)

Orlandoxx:


AbuBaniaz:
Very good step forward, but we can't merge this. You committed the po file in the wrong directory.  Can you move it to the `po` directory please? You can do that in the `fi.po changes` branch.

Orlandoxx:
Wrong directory.

PR #759, opened 2022-03-16
Change incorrect "Dobly Digital" to "Dolby digital"

BrianG61UK:


Huevos:
Why is this making changes to config.py ?

Why are you changing thing like "xxx.value = x" to "xxx.setValue(x)" ?

You can't mix up a pull request like this.

One pull request one change.

AbuBaniaz:
If you look at the files changed, it is just the message in infobargenrics and po files he wants changed.

We can get rid of noise commits when merging

Huevos:
If there is a PR it should only make one change.

And when working on a feature branch it should be rebased so there are no merge commits.

BrianG61UK:
@Huevos 
> If there is a PR it should only make one change.
It did only do one thing. It changed every instance of "Dobly Digital" to "Dolby digital".
You've done a couple of them yourself now but ignored all the ones in po and pot files.

BrianG61UK:
WTH? How does rebase work?

BrianG61UK:
> If you look at the files changed, it is just the message in infobargenrics and po files he wants changed.
> 
> We can get rid of noise commits when merging

@Huevos can't read so he doesn't know how.

He's an out of control power mad tyrant.

PR #760, opened 2022-03-17
Finish correcting DoblyDigital typo

BrianG61UK:
Finish correcting DoblyDigital typo which @Huevos partially fixed in the same commit as he fixed audioSelectionLong


Huevos:
This does not match the string in the .py file.

`"Dolby Digital downmix is now %s"`

BrianG61UK:
@Huevos So what you're saying is that you made the mistake of not changing the po files (and pot file) because you're lazy rather than because you didn't realize it needed doing?
There is now clearly absolutely no way you could have thought they were already correct because you changed the string yourself!

Huevos:
@BrianG61UK this is a job for the translators. How would I know where "%s" should be inserted into the translated string in 30 languages.

BrianG61UK:
@Huevos pity you didn't point that out earlier when I first mentioned po and pot files on the forum.

PR #765, opened 2022-04-22
DO NOT USE

BrianG61UK:


AbuBaniaz:
Brian, I can remove the 9 out of 10 commits when merging. I.e leaving one commit. However, it won't satisfy some quarters to only have commit that way. I think it will be best you resubmit the pull request .

Can you also remove the Brian was here in the comment?

Huevos:
Sorry but I don't want to switch off caching.

BrianG61UK:
Oops sorry. I didn't mean to do this.
It wasn't intended as a serious suggestion to be included in OpenViX.
Apologies.

PR #775, opened 2022-06-01
Update InfoBar.py

sfeary:
Minor addition to allow the exit button to behave the same as the stop button and exit the media player. Needed from an accessibility perspective for a user with limited mobility.


AbuBaniaz:
What difference is there between this requested change and the existing `"Quit Movie Player with EXIT button"` option?

sfeary:
Hi Abu, sorry I should have been more descriptive in my summary. All options other than "No" will exit the player and drop you back to live television. This change plus the corresponding config setting change I submitted will allow you to mimic the stop button and bring you back to the recordings list by pressing exit.

Huevos:
These pull requests are in the wrong order.

If this is merged please do it in the correct order.

PR #777, opened 2022-06-17
Added 2160p30 mode

zukon:
Added 2160p30 mode


TwolDE2:
You already have that in automatic resolution options

zukon:
Not everybody likes automatic resolution, but OK.

PR #797, opened 2022-10-19
Ice tv merge py3

prl001:
Changes to support the IceTV plugin.

They may be useful for other applications.

Main changes:
- Add a timers.xml attribute to store IceTV timer ids (similar to AutoTimer ids).
- Add callback mechanisms to RecordTimer for callbacks on timer creation, modification and deletion.
- Try to fetch timer name and description from the EPG if they are missing when the timer is created.
- Allow timer creation to specify the event_id.
- Add genre tables for the Australian and (now suspended) German IceTV services. Allow both a single ServiceReference or a list, allow servicerefs to be either tuples or strings, allow either a single genre or a list of genres to be imported, add the ability to import a tuple of list of tuples representing age ratings of programs.
- Add a new GenreList conversion type to the EventName conversion to allow lists of program genres to be displayed.
- Fix bug that prevented lists of genres being returned from the C++ EPGCache code to Python.

### DIFF
--- a/lib/dvb/db.h
+++ b/lib/dvb/db.h
@@ -73,12 +73,12 @@ public:
 	RESULT getBouquet(const eServiceReference &ref, eBouquet* &bouquet);
 //////
 	void loadBouquet(const char *path);
-	eServiceReference searchReference(int tsid, int onid, int sid);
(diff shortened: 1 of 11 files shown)


prl001:
I have previously discussed these changes with Huevos.

PR #804, opened 2022-12-08
Updated finnish translation

Orlandoxx:
InfoBar delete "..." from HelpableActionMaps


Orlandoxx:
Fixed version coming up.

PR #821, opened 2023-01-24
Add back code lost in changes to Py3 and update the comment to reflec…

original-birdman:
…t these.


Huevos:
You need to state the encoding.

original-birdman:
> You need to state the encoding.

No you don't.  The documented default encoding is utf8, and that is what we are dealing with with Unicode string <-> byte string conversions.

Huevos:
I know what the default encoding is, but if that changes your code will be broken.

original-birdman:
If the default encoding changes then I would expect a large number of things to be broken.
This is why APIs are documented and not changed.

PR #822, opened 2023-01-27
Add back code lost in changes to Py3 and update the comment to reflec…

original-birdman:
……t this.


original-birdman:
Now with added encoding options.

PR #832, opened 2023-02-17
Update About.py

Orlandoxx:
Fix localization


Orlandoxx:
Thanx TwolDE2 and Huevos (for creating the fix).

PR #840, opened 2023-03-04
Remove README entry about site.conf

biot:
There is no site.conf to edit, so it's unclear where you would even create it. However it gets created automatically on the first build, and sets BB_NUMBER_THREADS, BB_NUMBER_THREADS and DL_DIR properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,24 +90,17 @@
 
 
 ----------
-13 - Update site.conf
-
-    - BB_NUMBER_THREADS, PARALLEL_MAKE set to number of threads supported by the CPU
-    - add/modify DL_DIR = " location for build sources " to point to a location where you can save derived build sources,
-    this will reduce build time in fetching these sources again.
-
-----------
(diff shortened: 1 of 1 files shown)


TwolDE2:
This is more of an awareness comment, in that once site.conf is created it requires (for most people) an edit, which is why people who build frequently keep a copy to update any new build environments.
In particular, the location of fetched files needs to be set to a saved location…. This comment at least alerts people to the fact that you need to do something.

biot:
Yes, but I'm saying it automatically gets set to exactly what the entry says: threads/parallel got set to 8 for me, and DL_DIR was set to a "build-environment/sources", which it created. So the entry is superfluous.

TwolDE2:
… but that is not always the best setting, and it really should be changed to save the sources.

Huevos:
It gets created when you init the build folder.

If you find this entry in the readme superfluous just skip it, but it is not being deleted.

PR #865, opened 2023-04-14
Updated finnish translation

Orlandoxx:
For example : [ServiceInfo] tweak audio language formating


Orlandoxx:
[ServiceInfo] tweak audio language formating change reverted

PR #876, opened 2023-06-24
[Satconfig] Reword "Connected"

AbuBaniaz:
Include wording that is in hint text


Huevos:
How does this look in the interface? Is it OK in skins that are frugal on space?

AbuBaniaz:
This is how it looks in Vix Night HD. 
connected though another tuner

This is Kiddac's Slyk Q
connected though another tuner slykq

Which frugal skin in particular are you concerned about? Some of the devices listed have similar number of characters to the proposed change,  so they would have a similar issue, if it exists.

PR #883, opened 2023-07-12
Make the boxtype test simpler to read and modify.

original-birdman:
Set a flag from the result for later use rather than redo the same test.


original-birdman:
This also _only_ stop the `libreader` if it started it, which make more sense (to me).

PR #894, opened 2023-09-18
Put the Audio Codec description edit into one place, and make it read…

original-birdman:
…able.


original-birdman:
The replaces on all 4 of the ~700 character lines were all the same.

PR #905, opened 2023-10-19
[eListboxPythonMultiContent] added missing font variable to the formu…

DimitarCC:
…la for calculating max text width

### DIFF
--- a/lib/gui/elistboxcontent.cpp
+++ b/lib/gui/elistboxcontent.cpp
@@ -959,10 +959,15 @@ int eListboxPythonMultiContent::getMaxItemTextWidth()
 				{
 				case TYPE_TEXT: // text
 				{
-					ePyObject px = PyTuple_GET_ITEM(item, 1), pstring = PyTuple_GET_ITEM(item, 7);
+					ePyObject px = PyTuple_GET_ITEM(item, 1), pfnt = PyTuple_GET_ITEM(item, 5), pstring = PyTuple_GET_ITEM(item, 7);
 
 					if (PyLong_Check(pstring) && data) /* if the string is in fact a number, it refers to the 'data' list. */
 						pstring = PyTuple_GetItem(data, PyLong_AsLong(pstring));
+
+					if (pfnt) {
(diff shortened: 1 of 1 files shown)


lib/gui/elistboxcontent.cpp
```diff
@@ -959,11 +959,16 @@ int eListboxPythonMultiContent::getMaxItemTextWidth()
 				{
 				case TYPE_TEXT: // text
 				{
-					ePyObject px = PyTuple_GET_ITEM(item, 1), pstring = PyTuple_GET_ITEM(item, 7);
+					ePyObject px = PyTuple_GET_ITEM(item, 1), pfnt = PyTuple_GET_ITEM(item, 5), pstring = PyTuple_GET_ITEM(item, 7);
 
 					if (PyLong_Check(pstring) && data) /* if the string is in fact a number, it refers to the 'data' list. */
 						pstring = PyTuple_GetItem(data, PyLong_AsLong(pstring));
 
+					if (pfnt) {
```
Taapat, line 967: ~~I think there should be an additional check if the following font is in the list:~~
`if (pfnt) && !(m_font.find(fnt) == m_font.end()) {`
https://github.com/OpenViX/enigma2/blob/Developer/lib/gui/elistboxcontent.cpp#L1168-L1172

lib/gui/elistboxcontent.cpp
```diff
@@ -959,11 +959,16 @@ int eListboxPythonMultiContent::getMaxItemTextWidth()
...
+						int fnt_i = PyLong_AsLong(pfnt);
+						fnt = m_font[fnt_i];
```
Taapat: Sorry, check is needed before this.
`if !(m_font.find(fnt_i) == m_font.end()) fnt = m_font[fnt_i];`

DimitarCC: In my opinion that can not happen since if that is the case it will crash much before that function to be executed.

But well a guard check can be added anyway. I have added it.

PR #924, opened 2023-11-14
ServiceList two rows alternative visual mode patches

DimitarCC:
[Added] Options handling for new service list visual mode
[Fixed] some alignments issues


AbuBaniaz:
Are you sure the 'noqa' bit needs removing?

DimitarCC:
Indeed i missed this. Added it back ;)

PR #989, opened 2024-02-16
[epgcache] fix OPENTV, ATSC_EIT, EPG_IMPORT source type

LraiZer:
OPENTV and ATSC_EIT are no longer managing their own cached epg since this commit here.

https://github.com/OpenViX/enigma2/commit/05ae1ffc0159a4dc4f5a0f1c3322c21fdd1e0d7a#diff-b15024bd2b4ea380025955f8d690ba180059716f8f64bd2366de9dcd0d44b0c0R1920

int source is not carried through this overload function, it is being reset as EPG_IMPORT source.

These sources currently use this broken overload function: eEPGCache::OPENTV
eEPGCache::ATSC_EIT

### DIFF
--- a/lib/dvb/epgcache.cpp
+++ b/lib/dvb/epgcache.cpp
@@ -1917,7 +1917,7 @@ void eEPGCache::submitEventData(const std::vector<int>& sids, const std::vector<
 	{
 		event_types.push_back(event_type);
 	}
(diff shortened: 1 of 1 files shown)


Huevos:
Thanks

LraiZer:
Same issue with OpenATV if you would like to inform them, must have been a code copy/paste oversight during patch creation.

PR #994, opened 2024-03-11
[EventName] fix ETSI Classifications

LraiZer:
ETSI Classifications for min age 18 years is wrongly converted to Rating defined by broadcaster. since this patch: https://github.com/OpenViX/enigma2/commit/9702fa89efeeb60a87851d0e349b3fe1dc2a366f

### DIFF
--- a/lib/python/Components/Converter/EventName.py
+++ b/lib/python/Components/Converter/EventName.py
@@ -31,7 +31,7 @@ class ETSIClassifications(dict):
 			return "ratings/ETSI-%d.png" % age
 
 	def __init__(self):
-		self.update([(i, (self.shortRating(c), self.longRating(c), self.imageRating(c))) for i, c in enumerate(range(0, 15))])
+		self.update([(i, (self.shortRating(c), self.longRating(c), self.imageRating(c))) for i, c in enumerate(range(0, 18))])
 
 
 class AusClassifications(dict):
--- a/lib/python/Components/Converter/EventName.py
+++ b/lib/python/Components/Converter/EventName.py
@@ -31,7 +31,7 @@ class ETSIClassifications(dict):
 			return "ratings/ETSI-%d.png" % age
 
 	def __init__(self):
-		self.update([(i, (self.shortRating(c), self.longRating(c), self.imageRating(c))) for i, c in enumerate(range(0, 18))])
+		self.update([(i, (self.shortRating(c), self.longRating(c), self.imageRating(c))) for i, c in enumerate(range(0, 16))])
 
 
 class AusClassifications(dict):


jbleyel:
This is totally wrong, imho.

Goto Page 49

https://www.etsi.org/deliver/etsi_i_ets/300400_300499/300468/02_30_113/ets_300468e02v.pdf

IanSav:
A number of changes being made to the EPG / EIT, ratings, etc are incorrect.

After receiving this commit notice I reviewed my work from 7 years ago and can confirm that my code is correct.  Ratings are stored in a 4 bit nibble so the only possible values are 0 to 15!  The reference posted by @jbleyel is correct.  For Australia the data comes from AS 4599.1:2015 on page 67 (from memory) and is also correct in the code.

I have already posted to @LraiZer in another thread that his ratings changes are also incorrect.  Some of the changes are based on movie and DVD ratings rather that digital terrestrial television standards.  I didn't check the genre data but I suggest caution and checking the references.

LraiZer:
Read it many times ;)
especially the part:
```
0x00 undefined
0x01 to 0x0F minimum age = rating + 3 years
0x10 to 0xFF defined by the broadcaster
EXAMPLE: 0x04 implies that end users should be at least 7 years old.
```
minimum age = rating + 3 years
18 = 0x0F + 3

How do you intend to fix this bug then, if you don't fix it by changing the enumerate age range to (0-18)? 

When broadcaster sends 0x0F, it should be interpreted in GUI as a return string "Minimum age 18 years", as defined in the standard you have read yourself now.

However, when 0x0F is sent, it currently wrongly displays "Rating defined by broadcaster - %d" because enumerate range only for the 4 bit nibble sent by broadcaster (0-15) and not the age.

Huevos:
I'm not familiar with this code, but...
```
	def shortRating(self, age):
		if age == 0:
			return _("All ages")
		elif age <= 15:
			age += 3
			return " %d+" % age

	def longRating(self, age):
		if age == 0:
			return _("Rating undefined")
		elif age <= 15:
			age += 3
			return _("Minimum age %d years") % age

	def imageRating(self, age):
		if age == 0:
			return "ratings/ETSI-ALL.png"
		elif age <= 15:
			age += 3
			return "ratings/ETSI-%d.png" % age
```
All those seem to return None if the input > 15. What am I not understanding?

Huevos:
Ok, for sure the original code is wrong, because `range(0, 15)` produces 0-14. That is why the 15 is missing.

```
>>> list(enumerate(range(0, 15)))
[(0, 0), (1, 1), (2, 2), (3, 3), (4, 4), (5, 5), (6, 6), (7, 7), (8, 8), (9, 9), (10, 10), (11, 11), (12, 12), (13, 13), (14, 14)]
>>>
```

LraiZer:
Ok, not 18 then change it to 16.. if 15 not found then it defaults to here:
https://github.com/OpenViX/enigma2/commit/9702fa89efeeb60a87851d0e349b3fe1dc2a366f#diff-038e01f0833a65315d1926dde68f3362367e29b1aec903e200d2d4640879c874R51-R57

Huevos:
Sorry if I'm not clear.

Here, introduced by @IanSav, https://github.com/OpenViX/enigma2/commit/9702fa89efeeb60a87851d0e349b3fe1dc2a366f#diff-038e01f0833a65315d1926dde68f3362367e29b1aec903e200d2d4640879c874R25.

He is using `range(15)`. Should be `range(16)`.

```
>>> for i in range(15):
...     print("0x%x" % i)
...
0x0
0x1
0x2
0x3
0x4
0x5
0x6
0x7
0x8
0x9
0xa
0xb
0xc
0xd
0xe
>>>
```
`range(15)` only outputs up to `0xe`, when `0xf` is required. 

So looks like the bug was present in the initial commit.

IanSav:
I accept the range should have been 16 but anything higher is undefined and outside of the specifications.

PR #996, opened 2024-03-15
EPGCache redesign

DimitarCC:
[Removed] dxNoEIT flag since not clear what it do
[Removed] blacklist.onid file handling
[Updated] Handling of EIT and epg events
[Added] blacklist for supressing EIT or NOWNEXT data 
[Added] Whitelist for forcefully replace the epg cache entries with those from EIT/NOWNEXT 
[Added] function available in python for reloading of black/whitelists

### DIFF
--- a/lib/dvb/epgcache.cpp
+++ b/lib/dvb/epgcache.cpp
@@ -415,13 +415,7 @@ eEPGCache::eEPGCache()
 	CONNECT(eDVBLocalTimeHandler::getInstance()->m_timeUpdated, eEPGCache::timeUpdated);
 	CONNECT(cleanTimer->timeout, eEPGCache::cleanLoop);
 
-	std::ifstream onid_file;
-	onid_file.open("/etc/enigma2/blacklist.onid");
-	int tmp_onid;
-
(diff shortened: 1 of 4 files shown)


AbuBaniaz:
The noEIT flag was added so that specific channels/services can be marked to not use EIT so that they use web based EPG.  This allows people to use EPG importer for some services and EIT for others.

DimitarCC:
Yes I know for what was intended to be used that dxNoEIT flag. But it actually not works or work in some cases unclear how.
Thats why i find it confusing and removed it. There is new functionality that will replace that flag.


Best Regards,
Dimitar Tsenev


------ Original Message ------
From "AbuBaniaz" ***@***.******@***.***>>
To "OpenViX/enigma2" ***@***.******@***.***>>
Cc "Dimitar Tsenev" ***@***.******@***.***>>; "Author" ***@***.******@***.***>>
Date 15.3.2024 г. 14:38:02
Subject Re: [OpenViX/enigma2] EPGCache redesign (PR #996)


The noEIT flag was added so that specific channels/services can be marked to not use EIT so that they use web based EPG. This allows people to use EPG importer for some services and EIT for others.

PR #1001, opened 2024-03-27
[Updated] readded some fixes that got lost. Thanks @Ev0-BH

DimitarCC:


Orlandoxx:
Is this better accepted to merge ?

PR #1014, opened 2024-04-09
Dynamic size for Main Menu

DimitarCC:
[Added] Possibility to use List source with Addons
[Added] New addon for Main menu rendering
[Added] new events in Screen and List


Orlandoxx:
Hello, 
MainMenu.py : there is a typo on line 85 : def setManWidth(self, value): --> def setMaxWidth(self, value):

DimitarCC:
Indeed. Fixed.

PR #1022, opened 2024-04-22
[genre.py] fix "OpenTV" unhandled error

LraiZer:
TypeError: argument of type 'NoneType' is not iterable


BrianG61UK:
@LraiZer Excellent. Many thanks.

PR #1055, opened 2024-07-17
[unicable] fix number 'ub' when close tuner

Dima73:
-add longer timeout(default - use lnb boot_up_time)  when set voltage


Huevos:
@Dima73 how do we trigger the problem? What boxes?

Dima73:
> @Dima73 how do we trigger the problem? What boxes?
https://forums.openpli.org/topic/16424-some-questions-bugs/?view=findpost&p=1619377

PR #1070, opened 2024-09-21
[Updated]  Adjust addons to respect new text blending possibilities

DimitarCC:


Orlandoxx:
Hello, how can you utilize this new text blending ?

DimitarCC:
For normal label widgets you can just set alphatest="blend". Just like for pixmaps but now works for labels too.

Orlandoxx:
Okay, thanx.

PR #1098, opened 2024-12-21
Ensure the was_controlled_shutdown flag is cleared even when enigma2 …

original-birdman:
…dies from a signal.

### DIFF
--- a/tools/enigma2.sh.in
+++ b/tools/enigma2.sh.in
@@ -344,6 +344,12 @@ fi
 
 wait ${enigma2pid}
 
+# If enigma2 exited from an OS signal the was_controlled_shutdown flag
+# will not have been cleared.
+# So ensure that it is.
+#
+sed -i '/^config\.usage\.power\.was_controlled_shutdown=/d' @sysconfdir@/enigma2/settings
+
 # enigma2 exit codes:
 #
 # 1 - halt


Huevos:
You don't seem to be working on the latest version.

Please pull and try again.

original-birdman:
I did a pull 5 mins before adding the fix.

As I switched back to the Developer branch after pushing my fix I get:
```
[parent]: git checkout Developer
Switched to branch 'Developer'
Your branch is up-to-date with 'upstream/Developer'.
```
So what am I behind?

Huevos:
You are doing the same thing that has already been done 25 lines above.

https://github.com/OpenViX/enigma2/blob/bbc770ee5939934a9a2821dab1137b915220985e/tools/enigma2.sh.in#L326-L351

original-birdman:
Ah, I see what you mean.
Although that other addition is in the wrong place, as it precedes the wait ${enigma2pid}, which is where the script waits after enigma2 starts. So that code wipes out the value as soon as enigma2 starts up.
And the sed command is simpler.

The "# Enable generation of core dumps with ELF header included." section is where it is for a reason.

So revert that one and add mine.

Huevos:
If I revert that your commit won't merge. So can you make a commit that removes that and replaces with yours please.

PR #1118, opened 2025-03-31
[OScamInfo.py] whitespace cleanup

atefganm:
[OScamInfo.py] whitespace cleanup


Huevos:
This is wrong. PEP8 says 2 spaces before an inline comment.

https://peps.python.org/pep-0008/#inline-comments

PR #1130, opened 2025-04-24
[Added] Posibility to render virtual keyboard without using backgroun…

DimitarCC:
…d graphics. Instead use background colors. For use on slow receivers without HW alphablending

[Added] Bulgarian to virtual keyboard
[Fixed] Potential crash in ServiceName converter

### DIFF
--- a/lib/python/Components/Converter/ServiceName.py
+++ b/lib/python/Components/Converter/ServiceName.py
@@ -76,11 +76,10 @@ class ServiceName(Converter):
 			if not service:
 				if self.source.info:
 					sref = hasattr(self.source, "serviceref") and self.source.serviceref
-					sref = sref or service
-					nref = resolveAlternate(sref)
+					nref = sref and resolveAlternate(sref)
 					if nref:
 						sref = nref
-					return sref.toString()
+					return sref and sref.toString()
 				refstr = info.getInfoString(iServiceInformation.sServiceref)
 				path = refstr and eServiceReference(refstr).getPath()
 				if path and fileExists("%s.meta" % path):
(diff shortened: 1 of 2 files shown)


IanSav:
I would like to add your Bulgarian option to my original version of the code.  Could you please tell me if there are any local options, variations or disputes about this keyboard layout?  In the past there have been objections to some layouts so I would like to know if there might be any repercussions if I use your additions.

Also, your comment about performance issues of the very few background images is surprising.  Can you please give me some sense of scale of the issue you have encountered?  Given that boxes are getting faster not slower and the use of the VirtualKeyboard is minimal I would only consider this change if its effect is significant.

DimitarCC:
No problem to use this layout. It have no special requirements.
Also appears to work fine on my side.

As for performance issues...
Well i have issues on dreambox dm9x0, ab pulse4k, dual4k and some others.
That receivers have no hardware accelerated alphablending so they chokes on the keyboard operation.
Actually for every key there are minimum 3 pngs and that is the problem.

IanSav:
I wasn't checking the functionality.  In the past, when I have created language mappings, I have been embroiled in arguments that the button allocations are not acceptable to all members of the community.  I wish to avoid more of those confrontations.

I have never received any performance complaints in the past.  I have a PULSe 4K in my test rig.  I will run some tests to confirm your observations.  If confirmed, I will look for a solution.  I do not favour parameter use and have been phasing out their usage but I have some other ideas.

DimitarCC:
Well the button allocation is according to the Bulgarian national standard layout for keyboard. So no point someone to complain.

PR #1160, opened 2025-09-03
Update providers.py

nzepg:
to allow epg for New Zealand


AbuBaniaz:
@nzepg 
It seems you are using 4 spaces instead of tabs. The convention in Enigma2 is to use tabs. Whichever is correct, we cannot have a mix of both styles/preferences in one file. Can you please change the spaces to tabs in your commit.

Maybe we should add 13e too.

nzepg:
@AbuBaniaz what is 13e? Satellite EPG?

AbuBaniaz:
Hot bird, 13 degrees east. Sky Italia also has OpenTV EPG. Tuning details are in Cross EPG downloader..

Please leave it for now. Let's wait for your one to be merged first.

PR #1163, opened 2025-09-05
NzeUpdate providers.py

nzepg:
Added the EPG Changes for NZ, also included draft for Hotbird, that can be developed further


nzepg:
@Huevos will you add this yourself, maybe I don't have enough experience to use GitHub to meet your needs?

Huevos:
Nothing to do with github. You are making random changes with no explanation why.

PR #1164, opened 2025-09-06
Update providers.py

nzepg:


Huevos:
Why do you keep making random changes to 28.2E. This code has been working for years. If you are really in NZ this satellite is below the horizon for you. Why do you think you need to change it?

nzepg:
@Huevos  I am not actually changing any settings on 28.2E just adding comments to it.

Also, I couldn't see any frequency that is published https://www.lyngsat.com/Astra-2E-2F-2G.html that matches 11778000

AbuBaniaz:
There are no longer any  viewable services on 11778 o. 28.2. That is why Lyngsat does not list it.

PR #1189, opened 2025-12-17
Fix typos and some terms in german translation

xcentaurix:


AbuBaniaz:
You shouldn't change `msgid ` entries as those are made from the python files.

xcentaurix:
oh, sorry... will fix that.

PR #1201, opened 2026-01-16
[SoftcamManager] add support for gbox softcam

xcentaurix:
to avoid "non-standard softcam" message.


TwolDE2:
why do you need gbox cam? would be better moving to Oscam as that is where all the new changes exist or coming 
Better to talk on forum rather than get into maybe sensitive  softcam discussions here

xcentaurix:
thanks.

PR #1215, opened 2026-02-05
[encoding] fix old python2 utf-8 encoding

xcentaurix:
the old python2 umlaut style produced incorrect encoding when processed by the updateallpo.sh script. the new python3 style works fine.

### DIFF
--- a/lib/python/Components/Converter/genre.py
+++ b/lib/python/Components/Converter/genre.py
@@ -540,12 +540,12 @@ class GenresDEUIceTV:
 		1: (
 			_('Kunst'),  # 0x10 remapped from 0x01
 			_('Kultur'),  # 0x11 remapped from 0x01
-			_('Kriminalit\xc3\xa4t'),  # 0x12 remapped from 0x01
+			_('Kriminalität'),  # 0x12 remapped from 0x01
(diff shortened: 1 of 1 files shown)


Huevos:
You have non-ascii characters in python and no encoding is declared.

xcentaurix:
which non-ascii chars are you referring to? 
i thought in python3 utf-8 is default and doesnt need to be declared.

Huevos:
Did it compile?

If you use non-ascii chars you should force the encoding.
`# -*- coding: utf-8 -*-`

xcentaurix:
i tested it with a test script and it works:
(venv) (chroot) root@dm900:~/git$ cat test.py
t = "Tragödie"
print(t)
(venv) (chroot) root@dm900:~/git$ python test.py
Tragödie
(venv) (chroot) root@dm900:~/git$

in python3 you dont need any
# -*- coding: utf-8 -*-

btw. german umlauts are ascii-chars as well.

Huevos:
No, they are not ascii. Ascii only covers the first 127 chars.

xcentaurix:
anyway, the enigma2 build in the pr checks was fine.

PR #1227, opened 2026-02-14
[UsageCnofig] add date/time format string for skin

xcentaurix:
dummy strings for translation purpose only. will create entry in po files for translation.


Huevos:
Done slightly different inside InitUsageConfig(). https://github.com/OpenViX/enigma2/commit/6aa690529e4242a5f84d47e5636063c26af4a300

PR #1232, opened 2026-02-26
Re-enable stream corruption recovery.

Najar1991:
Removed the line that disabled the recovery handler. This allows the existing recovery logic to run when stream corruption is detected.


Huevos:
Can you explain exactly what this is and why it is needed.

Najar1991:
> Can you explain exactly what this is and why it is needed.

A special adjustment for time shift to increase stability when the channel code is interrupted to prevent drifft

PR #1250, opened 2026-04-04
Fix session handling and subtitle list injection issues

DimitarCC:
This pull request improves the handling of subtitles in the codebase by making the subtitle-related methods more robust to missing or unavailable service and session objects. The changes ensure that subtitle operations do not proceed when required dependencies are not available, preventing potential errors or crashes.

Error handling and robustness improvements:

* Added a check in `subtitleListInject` (in `InfoBar.py`) to ensure that the current service is available before attempting to find related subtitle files, and updated the method to use the current service from the session instead of a potentially outdated reference.
* Updated `_check_pts_and_show_sub` (in `SubtitleRenderer.py`) to verify that the `player` object has a valid `session` before proceeding, stopping subtitles and returning early if the session is missing.


Orlandoxx:
Thank you.

PR #1259, opened 2026-05-04
[Harddisk.py] Improve USB device mounting.

AbuBaniaz:


AbuBaniaz:
Duo 4K before
mount duo4k original

Duo 4K after
mount duo4k modded

DM900 Before
mount dm900 original

DM900 after
mount dm900 modded

TwolDE2:
Just use hotplug ot mountmanager…. Not everybody wants it mounted this way

AbuBaniaz:
## Improve USB device mounting

### Problem

On boxes with multiple storage devices, `sd*` USB/SATA partitions in `enumerateBlockDevices` were not being assigned correct mountpoints. They used whatever `getMountpoint()` returned from `/proc/mounts`, which on a restart contained stale mounts from the previous boot (e.g. `/media/sdb1`). Users had to manually configure mountpoints via MountManager.

### Solution

This PR improves the `elif partition.startswith("sd"):` block added in commit b26d953 to correctly handle stale kernel mounts and multiple devices, using only `self.partitions` (current session assignments) rather than `/proc/mounts` to determine which mountpoints are already in use.

### Changes

- Uses `self.partitions` instead of `/proc/mounts` to track used mountpoints, avoiding stale kernel mount interference
- Respects any existing fstab entry (UUID or `/dev/`-based, whether user-configured or written by `enumerateHotPlugDevices`) — if found, the device is left completely alone
- Remounts directly via `mount /dev/device /media/target` rather than writing fstab and calling `mount -a`, preventing stale entries accumulating across reboots
- Unmounts all stale `/media/` instances of the device before remounting, using mountpoint paths rather than the device path to correctly handle multiple mounts of the same device

### Behaviour

- First USB/SATA device → `/media/hdd`
- Additional USB/SATA devices → `/media/usb`, `/media/usb2` etc.
- Any existing fstab entry is always respected and never overwritten

### Notes

- Tested on dm900 with internal HDD and two USB drives across multiple reboots and enigma2 restarts

Huevos:
> ## Improve USB device mounting
> ### Problem
> 
> On boxes with multiple storage devices, `sd*` USB/SATA partitions in `enumerateBlockDevices` were not being assigned correct mountpoints. They used whatever `getMountpoint()` returned from `/proc/mounts`, which on a restart contained stale mounts from the previous boot (e.g. `/media/sdb1`). Users had to manually configure mountpoints via MountManager.
> ### Solution
> 
> This PR improves the `elif partition.startswith("sd"):` block added in commit [b26d953](https://github.com/OpenViX/enigma2/commit/b26d953b236fae34b43c819c2683c48f90f8f9db) to correctly handle stale kernel mounts and multiple devices, using only `self.partitions` (current session assignments) rather than `/proc/mounts` to determine which mountpoints are already in use.
> ### Changes
> 
>     * Uses `self.partitions` instead of `/proc/mounts` to track used mountpoints, avoiding stale kernel mount interference
> 
>     * Respects any existing fstab entry (UUID or `/dev/`-based, whether user-configured or written by `enumerateHotPlugDevices`) — if found, the device is left completely alone
> 
>     * Remounts directly via `mount /dev/device /media/target` rather than writing fstab and calling `mount -a`, preventing stale entries accumulating across reboots
> 
>     * Unmounts all stale `/media/` instances of the device before remounting, using mountpoint paths rather than the device path to correctly handle multiple mounts of the same device
> 
> 
> ### Behaviour
> 
>     * First USB/SATA device → `/media/hdd`
> 
>     * Additional USB/SATA devices → `/media/usb`, `/media/usb2` etc.
> 
>     * Any existing fstab entry is always respected and never overwritten
> 
> 
> ### Notes
> 
>     * Tested on dm900 with internal HDD and two USB drives across multiple reboots and enigma2 restarts

Is this a question? It seems to be an AI commentary, about what I don't know.

TwolDE2:
The changes in commit b26d953 were added to handle the dm9x0 SD card and also make use of what was then an unused large partition for data storage.
Subsequently, changes have been made to ensure that the SD card can be defined as /media/mmc....
Your request can be handled by correct use of hotplug (by removing and adding as different mount) or by MountManager, both which will also handle and update /proc/mounts(created by the linux kernel)  by issuing the correct mount/umount command.
Personally, I see no need to automatically add a device as /media/usb, preferring to give the user the option (I usually use /media/sd....  and this situation has been true for many years.

PR #1267, opened 2026-06-05
[enigma2] fix text cutoff in templated lists with RT_HALIGN_RIGHT

xcentaurix:
in templated lists last field in a row is cutoff by 10-20 pixels with scrollbar and without when using RT_HALIGN_RIGHT.
changing size or position of the field doesn't fix it.
problem with scrollbar (without is the same):
cutoffwithscrollbar

fix with scrollbar:
fixedcutoffwithscrollbar

fix without scrollbar:
fixedcutoffwithoutscrollbar


xcentaurix:
closing this pr after i learned that you can workaround the problem by tuning the field size of the last row element.

PR #1281, opened 2026-07-04
[enigma2] in ServiceOrbitalPosition converter return blank if service is not a satellite channel

xcentaurix:


xcentaurix:
replaced with a more bullet proof version: https://github.com/OpenViX/enigma2/pull/1282

PR #1284, opened 2026-07-06
[enigma2] Prevent main-thread hang from dead network-mounted picon paths

xcentaurix:
PiconLocator.__onMountpointAdded() now skips /media/net and /media/autofs mounts before adding them to searchPaths, mirroring the existing guard already in addSearchPath(). Previously, a network share that went unreachable after being mounted (server down, dropped connection) would leave a dead path in searchPaths, and findPicon()'s pathExists() check against it could block in uninterruptible sleep — freezing the entire reactor/main thread, since picon lookups run synchronously on it.


Huevos:
Merged

PR #1292, opened 2026-07-31
[PowerTimer] fix skin name resolution

xcentaurix:
missing skinName definition resulted in wrong name resolution to TimerEditListSummary instead of PowerTimerEditListSummary which results in incomplete skin display.
self.skinName = ["PowerTimerEditListSummary", "ScreenSummary"] fixes it.


xcentaurix:
fixed with this commit: https://github.com/OpenViX/enigma2/commit/feaf0cac23cdfceaf8e33aa83dd5f8a2fa333452

PR #1323, opened 2026-09-18
[eServiceMP3] Fix hang when switching embedded subtitles

Willo3092:
I have been investigating a reproducible hang when switching between embedded text subtitle tracks during MKV playback.

Testing was carried out on a clean OpenViX 6.9.002 installation on a Vu+ Uno 4K SE, using an MKV stored on the local SATA SSD with two embedded UTF-8 subtitle tracks. The first subtitle track enables normally, but switching to the other track causes playback to hang and the spinner to appear. The debug log stops at eServiceMP3 Clear Buffers!.

Reordering the subtitle tracks in MKVToolNix makes no difference — whichever track is selected first works, and switching to the other causes the hang. I also reproduced the same behaviour using the native eServiceMP3 player on OpenATV. The same file and subtitle tracks switch correctly using ServiceExtEplayer3.

I built the OpenViX 6.9.002 Enigma2 source corresponding to commit d3f089af4e06d60d0f3f8a7645331a8b31a65641 and tested it with only the clearBuffers() call in eServiceMP3::enableSubtitles() removed.

With that change I can repeatedly switch between both embedded subtitle tracks without a hang. I have also tested seeking forwards and backwards during playback and have not found any regression.

AI assistance was used during the investigation and build/test process.


TwolDE2:
So I talked fo Dimitar who added much of this code recently and he feels losing the clearbuffers causes other issues.
Maybe on the forum you can post him the mkv so that he can run tests?

